PR #9, opened 2020-03-30
KO-344 Move initcontainer functionality into main go cmd

eldondevat:
This moves some of the bash into go, modifies relevant yamls, improves some
logging, and makes the webhook ignore failure, in a scenario where a customer
has failed to configure it properly.

Rather than using the kubernetes CA, it creates a self-signed CA. This removes potential points of failure if certain permissions are not available in a given kubernetes deployment


operator/cmd/manager/main.go
```diff
@@ -142,26 +158,10 @@ func main() {
 		os.Exit(1)
 	}
 
-	skipWebhookEnvVal := os.Getenv("SKIP_VALIDATING_WEBHOOK")
```
eldondevat, line 145: From a different PR
@jimdickinson : can we keep this as a thing? not to use it by default, but as an escape hatch?

 @respringer: Yeah, I was a bit nervous about removing this as well. I recommend we keep it around at least until we have done one release with the webhook being defaulted.

jimdickinson:
Turned into https://github.com/datastax/cass-operator/pull/23

PR #19, opened 2020-03-31
ko-356 Add integration test for user provided superuser secret

johntrimble:


tests/testdata/default-single-rack-2-node-dc-with-superuser-secret.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: cassandra.datastax.com/v1beta1
+kind: CassandraDatacenter
+metadata:
+  name: dc2
+spec:
+  clusterName: cluster2
+  serverType: dse
+  serverVersion: "6.8.0"
+  serverImage: datastaxlabs/dse-k8s-server:6.8.0-20200316
+  configBuilderImage: datastaxlabs/dse-k8s-config-builder:0.9.0-20200316
```
jimdickinson, line 10: lets use datastax/cass-config-builder, and datastax/dse-server above

johntrimble: This is actually the same for all of our integration tests. If we want to update this for our tests, do you mind if I do that as part of the seed test since I have that one in progress?

PR #22, opened 2020-04-03
KO-365 minor reconcile refactor

sandoichi:


operator/pkg/reconciliation/context.go
```diff
@@ -37,6 +37,8 @@ type ReconciliationContext struct {
 	Services               []*corev1.Service
 	desiredRackInformation []*RackInformation
 	statefulSets           []*appsv1.StatefulSet
+	dcPods                 []*corev1.Pod
```
respringer, line 40: Can we have comments here to describe exactly what each field is?  Especially when we have two fields for something, like pods.

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -877,66 +789,53 @@ func isNodeStuckAfterLosingReadiness(pod *corev1.Pod) bool {
...
-		// FIXME make a enum, pods should start in an Init state
-		api.CassNodeState: stateStarted,
-	}
-	podList, err := listPods(rc, selector)
-	if err != nil {
-		rc.ReqLogger.Error(err, "no started pods found for CassandraDatacenter")
-		return false
-	}
+func (rc *ReconciliationContext) isClusterHealthy() bool {
+	pods := FilterPodListByCassNodeState(rc.clusterPods, stateStarted)
```
respringer, line 834: is it conventional for us to use "stateStarted" instead of STATE_STARTED ?

operator/pkg/reconciliation/context.go
sandoichi: This is actually a discussion point, because we (Jim and I) talked about it and weren't sure why we were differentiating between cluster pods and dc pods. Almost everything goes by dc pods, but 2 or 3 functions specifically used cluster pods. Ideally we can just use dcPods for everything if we don't have a good reason to use cluster pods.

operator/pkg/reconciliation/reconcile_racks.go
sandoichi: Camel case seems to be the golang convention for constants. If we already are using full caps in other places though, I can change it.

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1752,3 +1592,79 @@ func (rc *ReconciliationContext) CheckRollingRestart() result.ReconcileResult {
 
 	return result.Continue()
 }
+
+// Apply reconcileRacks determines if a rack needs to be reconciled.
```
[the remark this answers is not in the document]
jimdickinson: I like moving this function to the bottom of the file. Also, can you fix this comment?

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -547,19 +538,10 @@ func (rc *ReconciliationContext) CreateSuperuser() result.ReconcileResult {
...
-		api.ClusterLabel: rc.Datacenter.Spec.ClusterName,
-	}
-	podList, err := listPods(rc, selector)
-	if err != nil {
-		rc.ReqLogger.Error(err, "no pods found for CassandraDatacenter")
-		return result.Error(err)
-	}
-
-	pod := podList.Items[0]
+	pod := rc.clusterPods[0]
```
jimdickinson, line 541: this one can definitely be a dcPods reference without any problems

operator/pkg/reconciliation/reconcile_racks.go
jimdickinson: all caps is not a golang thing really, I spent time getting rid of that some months ago

operator/pkg/reconciliation/reconcile_racks_helpers.go
```diff
@@ -0,0 +1,90 @@
...
+	var endpoints = make(map[string]httphelper.EndpointState)
+	for _, ep := range metadata.Entity {
+		endpoints[ep.GetRpcAddress()] = ep
+	}
+	return endpoints
+}
+
+func PodPtrsFromPodList(podList *corev1.PodList) []*corev1.Pod {
+	var pods []*corev1.Pod
+	for idx, _ := range podList.Items {
```
jimdickinson, line 85: the linter will say this is equivalent

```suggestion
	for idx := range podList.Items {
```

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1659,32 +1479,25 @@ func isServerReady(pod *corev1.Pod) bool {
...
+		if err := rc.NodeMgmtClient.CallReloadSeedsEndpoint(pod); err != nil {
 			return err
 		}
 	}
 
 	return nil
 }
 
-func listPods(rc *ReconciliationContext, selector map[string]string) (*corev1.PodList, error) {
+func (rc *ReconciliationContext) listPods(selector map[string]string) (*corev1.PodList, error) {
```
jimdickinson, line 1500: 💯

operator/pkg/reconciliation/reconcile_racks_helpers.go
```diff
@@ -0,0 +1,90 @@
...
+func mapContains(base map[string]string, submap map[string]string) bool {
+	for k, v := range submap {
+		if val, ok := base[k]; !ok || val != v {
+			return false
+		}
+	}
+	return true
+}
+
+func FilterPodListByLabels(pods []*corev1.Pod, labelMap map[string]string) []*corev1.Pod {
```
jimdickinson, line 22: should probably add a docstring

operator/pkg/reconciliation/reconcile_racks_helpers.go
```diff
@@ -0,0 +1,90 @@
...
+	for _, p := range pods {
+		if mapContains(p.Labels, labelMap) {
+			filtered = append(filtered, p)
+		}
+	}
+	return filtered
+}
+
+func FilterPodListByLabel(pods []*corev1.Pod, labelName string, labelVal string) []*corev1.Pod {
+	filtered := []*corev1.Pod{}
```
jimdickinson, line 33: why not just forward this to `FilterPodListByLabels`?

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -418,7 +409,7 @@ func (rc *ReconciliationContext) CheckPodsReady() result.ReconcileResult {
 
 	// step 2 - get one node up per rack
 
-	rackWaitingForANode, err := rc.startOneNodePerRack(seedCount)
+	rackWaitingForANode, err := rc.startOneNodePerRack(endpointData, seedCount)
```
jimdickinson, line 412: I'm not sure about this change

sandoichi: How come? We have the potential to use the endpoint data in multiple checks (stuck pods, for example) so I extracted out to a single api call just like the podlist

operator/pkg/reconciliation/reconcile_racks_helpers.go
sandoichi: good point

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1333,11 +1225,10 @@ func (rc *ReconciliationContext) callNodeManagementStart(pod *corev1.Pod) error
 	return err
 }
 
-func (rc *ReconciliationContext) findStartingNodes(podList *corev1.PodList) (bool, error) {
+func (rc *ReconciliationContext) findStartingNodes() (bool, error) {
 	rc.ReqLogger.Info("reconcile_racks::findStartingNodes")
 
-	for idx := range podList.Items {
-		pod := &podList.Items[idx]
+	for _, pod := range rc.dcPods {
```
jimdickinson, line 1231: this could actually be a good use of cluster pods I think 😦

operator/pkg/reconciliation/reconcile_racks.go
jimdickinson: ok this makes more sense to me now

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1752,3 +1559,80 @@ func (rc *ReconciliationContext) CheckRollingRestart() result.ReconcileResult {
 
 	return result.Continue()
 }
+
+// ReconcileAllRacks determines if a rack needs to be reconciled.
+func (rc *ReconciliationContext) ReconcileAllRacks() (reconcile.Result, error) {
+	logger := rc.ReqLogger
+	logger.Info("reconcile_racks::Apply")
+
+	podList, err := rc.listPods(nil)
```
jimdickinson, line 1568: I'm not sure about this - maybe we should filter by pods the operator manages? I know you filter down right under this, but there's some potential this will bring back dozens/hundreds of other pods we aren't responsible for

sandoichi: actually what if we started off by passing in the cluster label filter here, so we initially get our clusterpods and then just grab the dc pods from there for the stuff that needs it. The dc filter already includes the cluster label, so it would cut down an extra filter step + avoid us grabbing anything that we didn't slap our cluster label on

operator/pkg/reconciliation/context.go
jimdickinson: yeah I don't think we can get rid of clusterPods, sadly

PR #23, opened 2020-04-03
KO-344 Move initcontainer functionality into main go cmd

eldondevat:
This moves some of the bash into go, modifies relevant yamls, improves some
logging, and makes the webhook ignore failure, in a scenario where a customer
has failed to configure it properly.

Rather than using the kubernetes CA, it creates a self-signed CA. This removes potential points of failure if certain permissions are not available in a given kubernetes deployment

> From #9 
> @jimdickinson : can we keep this as a thing? not to use it by default, but as an escape hatch?
> 
> @respringer: Yeah, I was a bit nervous about removing this as well. I recommend we keep it around at least until we have done one release with the webhook being defaulted.


operator/cmd/manager/main.go
```diff
@@ -117,11 +133,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err = ensureWebhookConfigVolume(cfg, namespace); err != nil {
+		log.Error(err, "Failed to ensure webhook volume")
+	}
+	if err = ensureWebhookCertificate(cfg, namespace); err != nil {
+		log.Error(err, "Failed to ensure webhook CA configuration")
```
jimdickinson, line 140: do we want to os.Exit(1) here like a lot of the scaffolded code does? and above too

tools/initcontainer/webhook_template.yaml
```diff
@@ -17,7 +17,7 @@ webhooks:
       path: /validate-cassandra-datastax-com-v1beta1-cassandradatacenter
     caBundle: AUTOMATICALLY_REPLACED_CA_BUNDLE
   admissionReviewVersions: ["v1beta1"]
-  failurePolicy: "Fail"
+  failurePolicy: "Ignore"
```
jimdickinson, line 20: 👍

hack/release/make-yaml-bundle.sh
```diff
@@ -21,7 +21,7 @@ echo '---' >> "$bundle"
 cat operator/deploy/crds/cassandra.datastax.com_cassandradatacenters_crd.yaml | yq r - >> "$bundle"
 
 echo '---' >> "$bundle"
-yq w operator/deploy/operator.yaml 'spec.template.spec.containers[0].image' 'datastax/cass-operator:1.0.0' >> "$bundle"
+yq w --doc 3 operator/deploy/operator.yaml 'spec.template.spec.containers[0].image' 'datastax/cass-operator:1.0.0' >> "$bundle"
```
jimdickinson, line 24: thanks for humoring me and looking into this!

operator/cmd/manager/main.go
eldondevat: I considered this but opted against it. If for some reason we are not successful here, the operator can still function effectively for many use cases.

I do think we should consider exposing some sort of health metric that we could put these kinds of exceptions on.

PR #27, opened 2020-04-07
Add helm chart

respringer:


charts/cass-operator-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Values.deploymentName }}
```
sandoichi, line 4: should these placeholders be named all specifically for cass operator, rather than leaving some of them generic? like `.Values.cassOperatorName` , `.Values.cassOperatorReplicas`, etc?

respringer: The only thing I named with a cassOperator prefix was cassOperatorNamespace, because there is a built-in NAMESPACE value that was conflicting with it.  As far as prefixing the other values, I can do it, but I don't think it's necessary.

See:

https://devcards.io/smurf-naming-convention

eldondevat:
Should we create mage/kind targets for this?

respringer:
> Should we create mage/kind targets for this?

I can if we want.  We could go so far as to refactor mage to always use helm internally for deploying and uninstalling the operator, but maybe that should be a follow-up PR if we choose to do it.

charts/cass-operator-chart/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+name: Cass Operator
+version: 1.0.0
+description: Helm chart for Cass Operator.
+maintainers:
+  - name: Ryan Springer
+    email: ryan.springer@datastax.com
```
johntrimble, line 7: I'm totally onboard with people emailing you when things don't work :-p

charts/cass-operator-chart/templates/deployment.yaml
johntrimble: It looks like elastic uses a generic "Namespace" placeholder in their helm chart:

https://github.com/elastic/helm-charts/blob/master/elasticsearch/templates/rolebinding.yaml#L19

Having said that, most of their resources do not seem to have a namespace set at all.

charts/cass-operator-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+    spec:
+      serviceAccountName: {{ .Values.serviceAccountName }}
+      volumes:
+      - name: cass-operator-certs-volume
+        secret:
+          secretName: cass-operator-webhook-config
+      containers:
+      - name: cass-operator
+        image: {{ .Values.cassOperatorImage }}
+        imagePullPolicy: {{ .Values.cassOperatorImagePullPolicy }}
```
johntrimble, line 24: Is the `cassOperator` prefix needed in this case?

README.md
```diff
@@ -31,6 +31,29 @@ NAME                             READY   STATUS    RESTARTS   AGE
 cass-operator-555577b9f8-zgx6j   1/1     Running   0          25h
 ```
 
+### (Optional) Loading the operator via Helm
```
jimdickinson, line 34: Can we push this new block down to directly above the `## Features` block? People have had trouble following our directions whenever anything was the slightest bit out of order, which any optional step kinda/sorta creates. Sorry!

In the longer form docs I'd be fine with it wherever.

charts/cass-operator-chart/templates/deployment.yaml
jimdickinson: yeah I think `namespace` might be a global thing for helm, kinda like it is for `kubectl`

charts/cass-operator-chart/templates/service.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
```
jimdickinson, line 3: this should have a namespace too

charts/cass-operator-chart/Chart.yaml
respringer: oh, I forgot to add everyone.... ;)

charts/cass-operator-chart/templates/deployment.yaml
respringer: Yeah, it's kinda wonky either way because the operator image is "cass-operator".  I'll go ahead and remove the prefix here.

README.md
respringer: No problem.

charts/cass-operator-chart/templates/service.yaml
respringer: Done.

charts/cass-operator-chart/templates/deployment.yaml
eldondevat: Regarding using a namespace, does it seem like this best practice is still valid? It _seems_ different from the approach we are using, but I may be misunderstanding:
https://github.com/helm/helm/issues/5465#issuecomment-473942223

charts/cass-operator-chart/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
```
eldondevat, line 1: I would be the helm v2 install base out there is pretty wide. Does using this api version preclude us from supporting those folks? https://helm.sh/docs/topics/charts/#the-apiversion-field

charts/cass-operator-chart/templates/deployment.yaml
respringer: There's a lot of back and forth in that thread, including a bunch of "confused" emojis on the comment you referenced.

Some people believe that this special-case treatment of namespace breaks configuration as code:

https://github.com/helm/helm/issues/5465#issuecomment-548247929

I think it breaks the "principal of least surprise" to treat namespace different than every other parameter, when it could just be handled by the default behavior.

charts/cass-operator-chart/Chart.yaml
respringer: I thought that we were intentionally targeting only Helm v2.  From what I understand, Helm v2 is a can of worms.

@jimdickinson ?

respringer:
If we want a mage-based integration test for the Helm chart, I would like to do that in a separate PR.

charts/cass-operator-chart/templates/deployment.yaml
eldondevat: https://github.com/helm/helm/issues/5465#issuecomment-573152021 I think the followup comment by the maintainer hits the point. They tried to make this change, and the community rejected it. There is a strong training effect with kubectl, where people are frequently specifying namespaces as arguments, so I think from that perspective "principal of least surprise" works on the side of specifying it as a helm argument. That said, I'm fine with your changes, and certainly not trying to block traffic here.

charts/cass-operator-chart/Chart.yaml
johntrimble: For what it's worth, among the people at KubeCon last year there were quite a few people that would have been quite happy to see Helm v2 burned to the ground, and pretty much everyone was glad to see Helm v3 get released. I don't know the technical details for all the angst with respect to v2, but I think the community has moved on.

jimdickinson: I thought the v3 binary was totally compatible with v2 charts?

charts/cass-operator-chart/templates/deployment.yaml
respringer: Yeah, if this is what is standard practice, then I am fine with changing it.  Does anyone feel strongly either way?

charts/cass-operator-chart/Chart.yaml
respringer: I will go and take a look, but if we intentionally release a chart that is compatible with Helm 2, then we are kind of saying that we support Helm 2.  Do we want to debug issues people might hit with Helm 2?

respringer: Helm 3 does support Helm api v1 ( that is, Helm 2 ) charts.  The three main differences I see:

1. instead of requirements.yaml there is Chart.yaml in Helm 3
2. Helm 2 charts do not support "library charts", which are useful for sharing values between multiple charts to keep things DRY.
3. Putting out a Helm 2 chart sends a message that we support Helm 2, and that doesn't seem like a great idea.  The Helm project itself really seems like it wants to move on and forget about Helm 2.

See: https://v3.helm.sh/docs/faq/#changes-since-helm-2

eldondevat: > I thought the v3 binary was totally compatible with v2 charts?

It does, but v2 of the charts is only for v3 of the binary, if I understand correctly.

> For what it's worth, among the people at KubeCon last year there were quite a few people that would have been quite happy to see Helm v2 burned to the ground, and pretty much everyone was glad to see Helm v3 get released. I don't know the technical details for all the angst with respect to v2, but I think the community has moved on.

I think that is totally true for new deployments, and I am one of those people who hates helm v2. If we don't have a good reason to support it, lets not! But I also know that customers have helmv2 out there, and might not be able to prioritize upgrading/reworking it.

charts/cass-operator-chart/templates/deployment.yaml
jimdickinson: Ok I researched this and it's very common to to use the generic helm namespace support...

https://cilium.readthedocs.io/en/stable/gettingstarted/k8s-install-gke/
https://istio.io/docs/setup/install/helm/#installation-steps
https://pingcap.com/docs/tidb-in-kubernetes/stable/deploy-tidb-from-kubernetes-gke/#deploy-tidb-operator

Linkerd adding a large explanation about why they couldn't support it, but it's confusing

https://linkerd.io/2/tasks/install-helm/#customizing-the-namespace

Mysql Operator made their own config 🤔 

https://github.com/oracle/mysql-operator/blob/master/docs/tutorial.md#configuration

README.md
```diff
@@ -129,6 +129,31 @@ UN  10.233.92.96    186.48 KiB  1            61.6%             b119eae5-2ff4-4b0
 UN  10.233.90.54    205.1 KiB   1            73.1%             0a96e814-dcf6-48b9-a2ca-663686c8a495  r1
 ```
 
+### (Optional) Loading the operator via Helm
+
+Helm may be used to load the operator.  The destination namespace must be created first.
```
jimdickinson, line 134: 💯 thank you for the changes wrt namespaces!

charts/cass-operator-chart/Chart.yaml
jimdickinson: From some research - I see other projects going with just Helm v3 support, so I think we can merge this PR and then deal with requests for Helm v2 support down the line.

PR #29, opened 2020-04-07
Add seed node update integration test

johntrimble:


tests/seed_selection/seed_selection_suite_test.go
```diff
@@ -0,0 +1,373 @@
...
+	// NOTE: The following check does not presently work due to 
+	// the lag time between when we update a seed label and when
+	// that change is reflected in DNS. Since we reload seed lists
+	// right after upating the label, some cassandra nodes will
+	// likely end up with slight out-of-date seed lists. KO-375
+	//
+	// checkCassandraSeedListsAlignWithSeedLabels(info)
+}
+
+func duplicate(value string, count int) string {
```
sandoichi, line 246: I'll take `Things that should have been included in the base libraries` for 100

sandoichi: Would probably be worthwhile to extract this out into the `ginkgo_util` package, since it could be used for any test with more than 1 node

tests/seed_selection/seed_selection_suite_test.go
```diff
@@ -0,0 +1,373 @@
...
+	result := []string{}
+	for i := 0; i < count; i++ {
+		result = append(result, value)
+	}
+
+	return strings.Join(result, " ")
+}
+
+func waitDatacenterReady() {
```
sandoichi, line 255: Again... it seems that you have written some nice things that could probably be ripped out into `ginkgo_util` and used in almost every test :thinking: 
I know there was some discussion at one point about how much these tests should just have the raw logic in them vs extracting things but it may be worth revisiting.

tests/testdata/default-single-rack-2-node-dc-with-superuser-secret.yaml
```diff
@@ -6,8 +6,6 @@ spec:
   clusterName: cluster2
   serverType: dse
   serverVersion: "6.8.0"
-  serverImage: datastaxlabs/dse-k8s-server:6.8.0-20200316
```
sandoichi, line 9: :+1:

tests/seed_selection/seed_selection_suite_test.go
johntrimble: Sounds like a good idea to me. I assumed `ginkgo_util` was for mage stuff, but looking at it I see that's not the case. We are already accumulating functions that are duplicated in multiple tests, so it's probably time to start extracting them.

johntrimble: Do you think this should be a method of `NsWrapper`?

sandoichi: Yeah probably...all of our tests are expected to run in a `test-` namespace so we would always be passing in the namespace anyway.

jimdickinson: Yes I wanted to extract this out too.

tests/seed_selection/seed_selection_suite_test.go
```diff
@@ -0,0 +1,373 @@
...
+	Ordinal int
+}
+
+type DatacenterInfo struct {
+	Size      int
+	RackNames []string
+	Nodes     []Node
+}
+
+func retrieveNodes() []Node {
```
respringer, line 82: At some point, we should look into moving some of this logic into our testing framework and share it between tests.

tests/seed_selection/seed_selection_suite_test.go
```diff
@@ -0,0 +1,373 @@
...
+				seeds = append(seeds, node.IP)
+			}
+			sort.Sort(sort.StringSlice(seeds))
+
+			Expect(seeds).To(Equal(expectedSeeds), "Expected pod %s to have seeds %v but had %v", node.Name, expectedSeeds, seeds)
+		}
+	}
+}
+
+func checkSeedConstraints() {
```
respringer, line 224: Would it be tacky to move this logic directly into the operator itself? ;)

tests/seed_selection/seed_selection_suite_test.go
respringer: I am huge fan of investing in a batteries-included testing lib.  The more simple tests are to write, the more likely we are to maintain a good amount of them.

PR #37, opened 2020-04-09
Add additional validation to the webhook.

respringer:


operator/pkg/apis/cassandra/v1beta1/webhook.go
```diff
@@ -5,35 +5,109 @@ package v1beta1
...
+		return fmt.Errorf("CassandraDatacenter attempted to change StorageConfig")
+	}
+
+	// Topology changes - Racks
+	// - Rack Name and Zone changes are disallowed.
+	// - Removing racks is not supported.
+	// - Reordering the rack list is not supported.
+	// - Any new racks must be added to the end of the current rack list.
+
+	if len(oldDc.Spec.Racks) > len(newDc.Spec.Racks) {
```
jimdickinson, line 64: sadly, we have to add the calls to GetRacks for both of these, I think

operator/pkg/apis/cassandra/v1beta1/webhook.go
```diff
@@ -5,35 +5,109 @@ package v1beta1
...
+	if dc.Spec.ServerType == "cassandra" && dc.Spec.ServerVersion != "3.11.6" && dc.Spec.ServerVersion != "4.0.0" {
+		return fmt.Errorf("CassandraDatacenter attempted to use unsupported Cassandra version '%s'",
+			dc.Spec.ServerVersion)
+	}
+
+	return nil
+}
+
+// Ensure that no values are improperly set
+func ValidateDatacenterFieldChanges(oldDc CassandraDatacenter, newDc CassandraDatacenter) error {
```
jimdickinson, line 35: great signature here and above

operator/pkg/apis/cassandra/v1beta1/webhook.go
respringer: Is this all to support the special case of the user not specifying any racks?

jimdickinson: that's right

Issue #26, opened 2020-04-07
update docs for expected keys for jvm options


jsanda:
I was trying to run one of the integration tests that was using this manifest:

```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: cluster1
  serverType: cassandra
  serverVersion: 3.11.6
  managementApiAuth:
    insecure: {}
  size: 3
  storageConfig:
      cassandraDataVolumeClaimSpec:
        storageClassName: server-storage
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 1Gi
  racks:
    - name: r1
    - name: r2
    - name: r3
  config:
    jvm-server-options:
      initial_heap_size: "800m"
      max_heap_size: "800m"
    cassandra-yaml:
      file_cache_size_in_mb: 100
      memtable_space_in_mb: 100
```
The config init container failed though with this exception:

```
$ kubectl logs cluster1-dc1-r1-sts-0 server-config-init
clojure.lang.ExceptionInfo: throw+: {:type :InvalidConfigKeys, :message "Invalid config keys for DSE version 3.11.6: jvm-server-options"} {:type :InvalidConfigKeys, :message "Invalid config keys for DSE version 3.11.6: jvm-server-options"}
	at slingshot.support$stack_trace.invoke(support.clj:201)
	at com.datastax.configbuilder.build_config$valid_config_keys_QMARK_.invokeStatic(build_config.clj:51)
	at com.datastax.configbuilder.build_config$valid_config_keys_QMARK_.invoke(build_config.clj:43)
	at com.datastax.configbuilder.build_config$build_configs.invokeStatic(build_config.clj:458)
	at com.datastax.configbuilder.build_config$build_configs.invoke(build_config.clj:450)
	at com.datastax.configbuilder.operator$make_configs.invokeStatic(operator.clj:36)
	at com.datastax.configbuilder.operator$make_configs.invoke(operator.clj:9)
	at com.datastax.configbuilder.operator$_main.invokeStatic(operator.clj:56)
	at com.datastax.configbuilder.operator$_main.doInvoke(operator.clj:52)
	at clojure.lang.RestFn.invoke(RestFn.java:397)
	at clojure.lang.AFn.applyToHelper(AFn.java:152)
	at clojure.lang.RestFn.applyTo(RestFn.java:132)
[...]

PR #38, opened 2020-04-09
document server jvm options

jsanda:
This PR is for #26.

The initial commit only covers a subset of the options for C* 3.11. I want to get some feedback before I do anything more. I am happy to help out with adding some docs however/wherever it is done.

I have always been a fan of Clojure. I even attended the very first Clojure Conj conference, but I definitely do not think user should have to hunt through Clojure code to find keys and their corresponding properties :)

### DIFF
--- a/docs/user/jvm_server_configuration.md
+++ b/docs/user/jvm_server_configuration.md
@@ -34,23 +34,25 @@ spec:
 ```
 The following table lists the supports keys that can appear under `jvm-options` along with their corresponding property names in `jvm.options`.
 
-| Config Builder key | jvm.options property | value type | notes | 
+| Config Builder key | jvm.options property | Value type | Notes | 
 | ------------------ | :-------------------:| :--------: | :---: |
+| `additional-jvm-opts` | `JVM_OPTS` | Arbitrary JVM options passed to Cassandra on start up |
 | `cassandra_ring_delay_ms` | `-Dcassandra.ring_delay_ms`| integer | Disabled by default |
 | `log_gc` | `-Xloggc:/var/log/cassandra/gc.log` | boolean | Disabled by default |
 | `thread_priority_policy_42` | `-XX:ThreadPriorityPolicy=42` | boolean | Enabled by default |
-| use_gc_log_file_rotation | `-XX:+UseGCLogFileRotation` | boolean | Disabled by default |
+| `use_gc_log_file_rotation` | `-XX:+UseGCLogFileRotation` | boolean | Disabled by default |
 | `initiating_heap_occupancy_percent` | `-XX:InitiatingHeapOccupancyPercent` | integer | Disabled by default. Can only be used when G1GC garbage collector is used. |
 | `string_table_size` | `-XX:StringTableSize` | string | Defaults to 1000003 |
 | `print_tenuring_distribution` | `-XX:+PrintTenuringDistribution` | boolean | Defaults to false |
 | `cassandra_initial_token` | `-Dcassandra.initial_token` | string | Disabled by default |
 | `resize_tlb` | `-XX:+ResizeTLAB` | boolean | Enabled by default |
 | `cassandra_join_ring` | `-Dcassandra.join_ring` | boolean | Enabled by default |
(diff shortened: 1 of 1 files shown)


jimdickinson:
this is awesome! I have a couple specific comments and I would like @cpmcdaniel to check it out before we merge it, but thank you so much for putting this together

docs/user/jvm_server_configuration.md
```diff
@@ -0,0 +1,61 @@
...
+| use_gc_log_file_rotation | `-XX:+UseGCLogFileRotation` | boolean | Disabled by default |
+| `initiating_heap_occupancy_percent` | `-XX:InitiatingHeapOccupancyPercent` | integer | Disabled by default. Can only be used when G1GC garbage collector is used. |
+| `string_table_size` | `-XX:StringTableSize` | string | Defaults to 1000003 |
+| `print_tenuring_distribution` | `-XX:+PrintTenuringDistribution` | boolean | Defaults to false |
+| `cassandra_initial_token` | `-Dcassandra.initial_token` | string | Disabled by default |
+| `resize_tlb` | `-XX:+ResizeTLAB` | boolean | Enabled by default |
+| `cassandra_join_ring` | `-Dcassandra.join_ring` | boolean | Enabled by default |
+| `use_tlb` | `-XX:+UseTLAB` | boolean | Enabled by default |
+| `perf_disable_shared_mem` | `-XX:+PerfDisableSharedMem` | boolean | Enabled by default |
+| `cassandra_config_directory` | `-Dcassandra.config` | string | Disabled by default |
```
jimdickinson, line 51: do we leave a note that using this is likely to break everything?

docs/user/jvm_server_configuration.md
```diff
@@ -0,0 +1,61 @@
...
+| `cms_wait_duration` | `-XX:CMSWaitDuration` | integer | Defaults to 10000. Can only be used when CMS garbage collector is used. |
+| `cassandra_replace_address` | `-Dcassandra.replace_address` | string | Disabled by default |
```
jimdickinson, line 53: this one too

cpmcdaniel:
I think this is great. Ideally I would like to maybe add some metadata fields in definitions so we can generate this type of documentation. There is already a `:description`, which should satisfy the last column. Perhaps a `:display-name` would be good for the `jvm.options property` column.

cpmcdaniel:
To be clear, that is not a recommendation for this PR. There is a lot of room for improvement on the documentation of config-builder input.

docs/user/jvm_server_configuration.md
jsanda: Definitely. I will update it.

docs/user/jvm_server_configuration.md
jsanda: Definitely. I will update it.

jsanda:
I have only taken an initial pass through the Cassandra 3.11 options so far. I was thinking about doing a table per server type / version.

I wonder if it would be better to have a single table for Cassandra and just add a column or note to point out properties that are specific to 3.11 or 4.0. Thoughts?

I will follow the same approach for DSE.

jsanda:
I just pushed changes for Cassandra 4.0. 

I am not sure how to document some of the JMX props like:

* `jmx-remote-ssl-opts`
* `jmx-remote-ssl`
* `jmx-remote-ssl-require-client-auth`

I also left out `jmx-port`. Is that another property that should have a warning?

I noticed that there are some [additional properties](https://github.com/datastax/cass-config-definitions/blob/f3f5e40c780100990729abc1c62498c520c26815/resources/jvm11-server-options/cassandra/jvm11-server-options-cassandra-4.0.0.edn) for Java 11. I have not documented these. Should I include them in this PR?

It would great to automate generating these docs. If that is going to happen, does it make sense for me to spend much more time on this?

If I do wind up spending more time on this, then I would prefer to add the DSE sections in a separate PR.

jimdickinson:
Yeah let's see about generating something like this before you spend more time on it. I like the way this looks for now, though!

PR #43, opened 2020-04-14
Events for mutations

johntrimble:
Here are the event types I added. Looking for feedback on the names to make the naming consistent:

```
$ grep Eventf operator/pkg/reconciliation//reconcile_racks.go | grep -oE '["][A-Za-z]+["][,]\s*$' | grep -Eo '[A-Za-z]+'
UpdatingRack
UpateRackLabelsOnResource
StoppingDatacenter
ScaleUpRack
CreateSuperuser
FinishReplaceNode
StartReplaceNode
DeletingStuckPod
SetPodAsSeed
UnsetPodAsSeed
CreatedResource
CreatedResource
UpateRackLabelsOnResource
UpateRackLabelsOnResource
StartedCassandra
StartCassandraAndReplaceNode
StartCassandra
SetPodAsSeed
RestartingCassandra
```

### DIFF
--- a/operator/pkg/reconciliation/reconcile_racks.go
+++ b/operator/pkg/reconciliation/reconcile_racks.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"sort"
 	"time"
+	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -194,6 +195,9 @@ func (rc *ReconciliationContext) CheckRackPodTemplate() result.ReconcileResult {
 
 		if needsUpdate {
 
+			rc.Recorder.Eventf(rc.Datacenter, "Normal", "UpdatingRack",
+				"Updating rack %s", rackName)
+
 			if err := setOperatorProgressStatus(rc, api.ProgressUpdating); err != nil {
 				return result.Error(err)
 			}
@@ -271,6 +275,9 @@ func (rc *ReconciliationContext) CheckRackLabels() result.ReconcileResult {
 				// FIXME we had not been passing this error up - why?
 				return result.Error(err)
 			}
+
+			rc.Recorder.Eventf(rc.Datacenter, "Normal", "UpateRackLabelsOnResource",
+				"Update rack labels for StatefulSet %s", statefulSet.Name)
 		}
 	}
 
@@ -280,6 +287,7 @@ func (rc *ReconciliationContext) CheckRackLabels() result.ReconcileResult {
 func (rc *ReconciliationContext) CheckRackStoppedState() result.ReconcileResult {
 	logger := rc.ReqLogger
 
+	emittedStoppingEvent := false
 	racksUpdated := false
 	for idx := range rc.desiredRackInformation {
 		rackInfo := rc.desiredRackInformation[idx]
@@ -294,6 +302,12 @@ func (rc *ReconciliationContext) CheckRackStoppedState() result.ReconcileResult 
 				"rack", rackInfo.RackName,
 				"currentSize", currentPodCount,
 			)
+
+			if !emittedStoppingEvent {
+				rc.Recorder.Eventf(rc.Datacenter, "Normal", "StoppingDatacenter",
+					"Stopping datacenter")
+				emittedStoppingEvent = true
+			}
 
 			rackPods := FilterPodListByLabels(rc.dcPods, rc.Datacenter.GetRackLabels(rackInfo.RackName))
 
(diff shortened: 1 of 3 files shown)


operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -194,6 +195,9 @@ func (rc *ReconciliationContext) CheckRackPodTemplate() result.ReconcileResult {
 
 		if needsUpdate {
 
+			rc.Recorder.Eventf(rc.Datacenter, "Normal", "UpdatingRack",
```
respringer, line 198: What does everyone think about defining an enum for the "reason" argument?  The docs say:

> // 'type' of this event, and can be one of Normal, Warning. New types could be added in future
> // 'reason' is the reason this event is generated. 'reason' should be short and unique; it
> // should be in UpperCamelCase format (starting with a capital letter). "reason" will be used
> // to automate handling of events, so imagine people writing switch statements to handle them.
> // You want to make that easy.

( https://godoc.org/k8s.io/client-go/tools/record#EventRecorder )

Making an enum for the reasons would allow us to easily audit all events at a high-level.

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -473,6 +487,9 @@ func (rc *ReconciliationContext) CheckRackScale() result.ReconcileResult {
 				"desiredSize", desiredNodeCount,
 			)
 
+			rc.Recorder.Eventf(rc.Datacenter, "ScaleUpRack",
```
respringer, line 490: According to https://godoc.org/k8s.io/client-go/tools/record#EventRecorder , the "type" argument can only be "Normal" or "Warning".

johntrimble: Err, yeah, that's a bug.

operator/pkg/reconciliation/reconcile_racks.go
johntrimble: You are right, we should make these constants. The reason values I have now are intended to be what is described in the docs. Do you want changes to the actual reason values?

respringer: No, the values seem appropriate to me.

respringer: If there is any feedback, it's just to be overly pedantic on verb tense.  If we have completely finished an action, make sure we use past tense like "CreatedResource", and if we are about to do something or we are currently doing it, let's make sure it's a continuous tense like "DeletingStuckPod".  The stuff like "SetPodAsSeed" is neither past tense nor continous, and is slightly ambiguous.

This is uber-pedantic, but it would remove the possibility of someone getting confused.

respringer: Also, "UpateRackLabelsOnResource" has a typo.  I broke the existing values into groups based on my grammar mania:

Ongoing Actions:

UpdatingRack
StoppingDatacenter
DeletingStuckPod
RestartingCassandra

Completed Actions:

CreatedResource
StartedCassandra
SetPodAsSeed
UnsetPodAsSeed

Ambiguous Actions:

UpateRackLabelsOnResource
ScaleUpRack
CreateSuperuser
FinishReplaceNode
StartReplaceNode
StartCassandraAndReplaceNode
StartCassandra

jimdickinson:
I brought up a cluster and my main comment is I would try to phrase the message so the variable part is always a suffix of the event...
```
- Making pod cluster1-dc1-rack2-sts-0 a seed node
+ Labeling seed node cluster1-dc1-rack2-sts-0
```

and I might shorten a name or two.

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1304,10 +1350,14 @@ func (rc *ReconciliationContext) startCassandra(endpointData httphelper.CassMeta
 		// If we have a replace address that means the cassandra node did
 		// join the ring previously and is marked for replacement, so we
 		// start it accordingly
+		rc.Recorder.Eventf(rc.Datacenter, "Normal", "StartCassandraAndReplaceNode",
```
jimdickinson, line 1353:
```suggestion
		rc.Recorder.Eventf(rc.Datacenter, "Normal", "ReplaceCassandraNode",
```

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1150,6 +1190,9 @@ func (rc *ReconciliationContext) ReconcilePods(statefulSet *appsv1.StatefulSet)
 					"PVC", pvc,
 				)
 			}
+
+			rc.Recorder.Eventf(rc.Datacenter, "Normal", "UpateRackLabelsOnResource",
```
jimdickinson, line 1194:
```suggestion
			rc.Recorder.Eventf(rc.Datacenter, "Normal", "LabeledRackResource",
```

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -824,6 +853,8 @@ func (rc *ReconciliationContext) deleteStuckNodes() (bool, error) {
 
 		if shouldDelete {
 			rc.ReqLogger.Info(fmt.Sprintf("Deleting stuck pod: %s. Reason: %s", pod.Name, reason))
+			rc.Recorder.Eventf(rc.Datacenter, "Normal", "DeletingStuckPod",
```
jimdickinson, line 856: 💯 this one is important

PR #46, opened 2020-04-16
Enable webhook by default

jimdickinson:
This PR flips our deployment yaml to bring up the webhook server by default. It ballooned in size when I wanted to update the helm chart and then use the helm chart to update our combined yaml manifest. Haven't touched tests here, but all of the int tests passed.


operator/deploy/cluster_role_binding.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: cass-operator
-  namespace: ""
+  namespace: "cass-operator"
```
jimdickinson, line 8: I thought this was a mistake when I was reviewing this, but it's just unclear what should be here

charts/cass-operator-chart/templates/clusterrole.yaml
```diff
@@ -4,42 +4,6 @@ metadata:
   creationTimestamp: null
   name: {{ .Values.clusterRoleName }}
 rules:
-- apiGroups:
```
respringer, line 7: +1 removing unnecessary cruft.

hack/release/make-yaml-bundle.sh
```diff
@@ -9,25 +9,7 @@ echo '---' >> "$bundle"
...
-
-echo '---' >> "$bundle"
-cat operator/deploy/role_binding.yaml | yq r - >> "$bundle"
-
-echo '---' >> "$bundle"
-cat operator/deploy/crds/cassandra.datastax.com_cassandradatacenters_crd.yaml | yq r - >> "$bundle"
-
-echo '---' >> "$bundle"
-yq w --doc 3 operator/deploy/operator.yaml 'spec.template.spec.containers[0].image' 'datastax/cass-operator:1.0.0' >> "$bundle"
+helm template ./charts/cass-operator-chart/ -n cass-operator | kubectl create --dry-run=client -o yaml -n cass-operator -f - >> "$bundle"
```
respringer, line 12: :) :) :)

operator/deploy/cluster_role_binding.yaml
respringer: FYI: This works for my integration test with a namespace of "".

hack/release/make-yaml-bundle.sh
sandoichi: :100:

operator/deploy/cluster_role_binding.yaml
jimdickinson: I think the code that needs this ClusterRoleBinding fails and continues - could be the explanation?

PR #48, opened 2020-04-20
Int test webhook

respringer:
Added a couple of tests to ensure that the webhook catches invalid updates in an integration test.


tests/testdata/cluster-role-binding_webhook-validation.yaml
```diff
@@ -0,0 +1,12 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cass-operator
+subjects:
+- kind: ServiceAccount
+  name: cass-operator
+  namespace: test-webhook-validation
```
respringer, line 8: @jimdickinson , actually I think I was mistaken in my namespace comment in the other PR.  I had to specify a value for namespace here to make things work.  It didn't accept "".  Therefore I don't think we can have a generic setting for namespace.  We have to rely on helm or some other process to set a value.

tests/webhook_validation/webhook_validation_suite_test.go
```diff
@@ -0,0 +1,113 @@
...
+				FormatOutput(json)
+			ns.WaitForOutputAndLog(step, k, "Ready", 30)
+
+			step = "attempt to use invalid dse version"
+			json = "{\"spec\": {\"serverType\": \"dse\", \"serverVersion\": \"4.8.0\"}}"
+			k = kubectl.PatchMerge(dcResource, json)
+			ns.ExecAndLogAndExpectErrorString(step, k,
+				"validation failure list:\nspec.serverVersion in body should be one of [6.8.0 3.11.6 4.0.0]")
+
+			step = "attempt to change the dc name"
```
respringer, line 95: The unit tests exercise the validation logic more thoroughly.

tests/testdata/cluster-role-binding_webhook-validation.yaml
jimdickinson: ok, this is actually good news, we just need a step to deal with it, and @sandoichi is working on that now

mage/sh/lib.go
```diff
@@ -53,6 +53,20 @@ func RunVPanic(cmd string, args ...string) {
...
 
+// Run command and print any output to stdout/stderr
+// Also return stdout/stderr as strings
+func RunVCapture(cmd string, args ...string) (string, string, error) {
+	captureOut := new(bytes.Buffer)
+	captureErr := new(bytes.Buffer)
+
+	// Duplicate the output/error to our buffer and the test stdout/stderr
+	multiOut := io.MultiWriter(captureOut, os.Stdout)
+	multiErr := io.MultiWriter(captureErr, os.Stderr)
```
jimdickinson, line 64: very cool 👍

PR #49, opened 2020-04-20
KO-389 use helm in integ tests and mage setup targets

sandoichi:


mage/ginkgo/lib.go
```diff
@@ -103,11 +102,11 @@ func (k *NsWrapper) countStep() int {
 	return n
 }
 
-func (k NsWrapper) Terminate() error {
+func (ns NsWrapper) Terminate() {
```
jimdickinson, line 105: why stop returning the error?

mage/helm/lib.go
```diff
@@ -0,0 +1,43 @@
...
+		fmt.Sprintf("--namespace=%s", namespace),
+	}
+
+	if overrides != nil && len(overrides) > 0 {
+		var overrideString = ""
+		for key, val := range overrides {
+			if overrideString == "" {
+				overrideString = fmt.Sprintf("%s=%s", key, val)
+			} else {
+
```
respringer, line 24: pedantic tip: remove this extra newline.

mage/kind/lib.go
```diff
@@ -156,27 +162,25 @@ func RunIntegTests() {
 // Perform all the steps to stand up an example Kind cluster,
 // except for applying the final cassandra yaml specification.
 // This must either be applied manually or by calling SetupCassandraCluster
-// or SetupDCECluster
+// or SetupDCECluster.
+// This target assumes that helm is installed and available on path.
 func SetupExampleCluster() {
 	mg.Deps(SetupEmptyCluster)
 	kubectl.CreateSecretLiteral("cassandra-superuser-secret", "devuser", "devpass").ExecVPanic()
-	kubectl.ApplyFiles(
```
respringer, line 163: :)

mage/ginkgo/lib.go
respringer: It looks like the return value is being ignored at least most of the time by the callers.  Still, maybe we should add panics at the caller instead of ignoring things.  I am not sure what is the best approach here.  What is the consequence of an error during Terminate() ?

tests/config_change/config_change_suite_test.go
```diff
@@ -50,18 +44,13 @@ var _ = Describe(testName, func() {
...
 
-			step := "creating default resources"
-			k := kubectl.ApplyFiles(defaultResources...)
-			ns.ExecAndLog(step, k)
-
-			step = "creating the cass-operator resource"
-			k = kubectl.ApplyFiles(operatorYaml)
-			ns.ExecAndLog(step, k)
+			step := "setting up cass-operator resources via helm chart"
+			ns.HelmInstall("../../charts/cass-operator-chart")
```
respringer, line 48: #winning

mage/ginkgo/lib.go
sandoichi: Yeah this is a tricky one. In general the functions here in the ginko util are in charge of doing a ginko friendly panic (usually `Expect`) as a way to have to put in less error handling logic in the tests themselves. 

As Ryan mentioned, we were always ignoring the return value of this, so it seemed weird to keep returning it. I'm not sure what the best way to handle the errors here is. We could query the cluster before each delete call here to make sure that a resource exists before attempting to delete, in case that causes an error that we don't care about (I think it does?). We also don't want to panic immediately, because we want to try and clean up as much as we can (probably?). 

We could perhaps keep a list of errors that occurred and at the end do something with them. Failing to cleanup is a bigger deal now with helm since subsequent runs will just fail out completely if any resources remain, and being in a partially cleaned up state will also prevent us from being able to do a `helm uninstall` on a subsequent run to resolve it.

jimdickinson: yeah I noticed we have a more graceful uninstall... maybe give this function a comment like it's the shutdown handler, and make sure the graceful one is error checking?

my concern is sometimes not fully cleaning up leaks resources that cost $$$, like PVs

mage/helm/lib.go
```diff
@@ -0,0 +1,43 @@
+// Copyright DataStax, Inc.
+// Please see the included license file for details.
+
+package helm_util
```
jimdickinson, line 4: just noticed the dirs mismatch the package statement - how'd we get in that pattern for mage?

mage/integ-tests/lib.go
```diff
@@ -68,13 +68,18 @@ func runGinkgoTestSuite(path string) {
...
+// Example:
+// M_INTEG_DIR=scale_up,stop_resume
+//
+// This target assumes that helm is installed and available on path.
 func Run() {
 	var testDirs []string
 	integDir := os.Getenv(envIntegDir)
 	if integDir != "" {
-		testDirs = []string{integDir}
+		testDirs = strings.Split(integDir, ",")
```
jimdickinson, line 82: 😻

mage/kind/lib.go
```diff
@@ -156,27 +162,25 @@ func RunIntegTests() {
 // Perform all the steps to stand up an example Kind cluster,
 // except for applying the final cassandra yaml specification.
 // This must either be applied manually or by calling SetupCassandraCluster
-// or SetupDCECluster
+// or SetupDCECluster.
```
jimdickinson, line 165:
```suggestion
// or SetupDSECluster.
```

mage/helm/lib.go
sandoichi: yeah we have this in most of our mage dirs. It probably started back with `util` and `sh` dirs being self explanatory when viewing in the filesystem, but when importing as just `sh` and `util` they seemed ambiguous and I think we changed the package names to compensate, but left the dirs more concise.

mage/ginkgo/lib.go
sandoichi: Helm will uninstall everything except for datacenters, crd, namespace itself. If any of helm's installed resources are not there anymore, it will fail to uninstall everything.

So helm is really only one chunk of the important cleanup here. It seems like we should probably track if any of the steps failed, cleanup as much as we can, and then report any errors?

```
1) Delete datacenters and store any errors
2) Helm uninstall, and store any errors
3) Delete crd
4) Delete namespace, and store any errors <--- though this step will probably still hang indefinitely if step 1 failed...maybe we shouldn't attempt it if step 1 failed.
```

jimdickinson: ^ looks right to me - but if it's gonna be more than a few more lines of changes you can punt it to another PR

PR #50, opened 2020-04-21
Add conditions

johntrimble:


operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1682,6 +1710,8 @@ func (rc *ReconciliationContext) ReconcileAllRacks() (reconcile.Result, error) {
 		return recResult.Output()
 	}
 
+	// TODO: Set condition Ready
+	
 	if err := setOperatorProgressStatus(rc, api.ProgressReady); err != nil {
 		return result.Error(err).Output()
 	}
```
respringer, line 1717: This file is getting uncomfortably large.  In a different ticket, we should consider splitting it up a bit.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -200,9 +183,30 @@ type CassandraNodeStatus struct {
...
+	DatacenterScalingUp      DatacenterConditionType = "ScalingUp"
+	DatacenterUpdating       DatacenterConditionType = "Updating"
+	DatacenterStopped        DatacenterConditionType = "Stopped"
+	DatacenterResuming       DatacenterConditionType = "Resuming"
+	DatacenterRollingRestart DatacenterConditionType = "RollingRestart"
+)
+
+type DatacenterCondition struct {
+	Type DatacenterConditionType `json:"type"`
+	Status corev1.ConditionStatus `json:"status"`
```
jimdickinson, line 201: 💯

mage/ginkgo/lib.go
```diff
@@ -213,6 +213,16 @@ func (ns *NsWrapper) WaitForOutputContainsAndLog(description string, kcmd kubect
 	Expect(execErr).ToNot(HaveOccurred())
 }
 
+
+func (ns *NsWrapper) WaitForDatacenterCondition(dcName string, conditionType string, value string) {
```
sandoichi, line 217: :+1:

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -200,9 +183,30 @@ type CassandraNodeStatus struct {
...
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
```
jimdickinson, line 202: did you ever figure out what was wrong with updating the slice? if you had a zero value here, that could be part of the issue, like here https://github.com/coreos/prometheus-operator/issues/2399

operator/pkg/events/events.go
```diff
@@ -0,0 +1,20 @@
+package events
```
sandoichi, line 1: :+1:

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1609,6 +1690,91 @@ func (rc *ReconciliationContext) CheckRollingRestart() result.ReconcileResult {
 	return result.Continue()
 }
 
+func (rc *ReconciliationContext) MaybeUpdateCondition(condition api.DatacenterCondition) bool {
```
jimdickinson, line 1693: how about just `setCondition()`?

jimdickinson, line 1693: I think this Type and Status should be the args here, it would remove a couple lines from each call-site

johntrimble: Because it is conditional? We already have a `SetCondition()` method on the Datacenter resource.

johntrimble: There could potentially be more fields on the `Condition` struct in the future. For example, a human readable string explaining why the condition was set. Some fields may be optional. Is it still worth putting the fields in the function signature?

jimdickinson: but I think it's fine that it's conditional - just add a docstring and keep the return value the same, no need for the `maybe`

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
johntrimble: I never did figure out what was going on with updating the slice, I actually need to check to see if it is still necessary to recreate the array. That timestamp was added after dealing with those issues though. I haven't seen this cause a problem yet, possibly because it never ends up being null?

operator/pkg/reconciliation/reconcile_racks.go
jimdickinson: hmm maybe keep the single arg the same, then, but add a convenience constructor `new DatacenterCondition()` now, and `newDatacenterConditionWithReason()` later

what do you think?

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1609,6 +1690,91 @@ func (rc *ReconciliationContext) CheckRollingRestart() result.ReconcileResult {
...
+	dc := rc.Datacenter
+	if dc.GetConditionStatus(condition.Type) != condition.Status {
+		// We are changing the status, so record the transition time
+		condition.LastTransitionTime = metav1.Now()
```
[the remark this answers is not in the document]
johntrimble: I think SetCondition is a lower level function. It's purpose is to take state and make sure it gets set. What the timestamp should be is I think a higher level concern. For example, an argument could be made that if multiple conditions get set simultaneously, then the timestamp should be the same for all of them.

operator/pkg/reconciliation/reconcile_racks.go
johntrimble: I don't think the size of the file is the real problem. It's the manner in which the functions are all interconnected.

operator/pkg/reconciliation/reconcile_racks.go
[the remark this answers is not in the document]
johntrimble: In other words, `SetCondition` is about how to set the state, not what to set it to.

operator/pkg/reconciliation/reconcile_racks.go
johntrimble: so `newDatacenterCondition()` takes the type and status as parameters? That works for me!

johntrimble: What about the similar names between the `SetCondition` on the Datacenter and the proposed `setCondition` method here? Is that confusing?

jimdickinson: oh I also suggested moving away from the word Update because we're not using the k8s client's Update function here

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
johntrimble: Just tested this. It works fine without creating a new array.

tests/rolling_restart/rolling_restart_suite_test.go
```diff
@@ -0,0 +1,84 @@
...
+			step = "trigger restart"
+			json := `{"spec": {"rollingRestartRequested": true}}`
+			k = kubectl.PatchMerge(dcResource, json)
+			ns.ExecAndLog(step, k)
+
+			// Ensure we actually set the condition
+			ns.WaitForDatacenterCondition(dcName, "RollingRestart", string(corev1.ConditionTrue))
+
+			// Ensure we actually unset the condition
+			ns.WaitForDatacenterCondition(dcName, "RollingRestart", string(corev1.ConditionFalse))
```
jimdickinson, line 68: this is great, very clean and captures the essence of the rolling restart, wave in the 🌊

tests/config_change_condition/config_change_condition_suite_test.go
```diff
@@ -0,0 +1,69 @@
...
+			json := "{\"spec\": {\"config\": {\"cassandra-yaml\": {\"file_cache_size_in_mb\": 123}, \"jvm-server-options\": {\"garbage_collector\": \"CMS\"}}}}"
+			k = kubectl.PatchMerge(dcResource, json)
+			ns.ExecAndLog(step, k)
+
+			ns.WaitForDatacenterCondition(dcName, "Updating", string(corev1.ConditionTrue))
+			ns.WaitForDatacenterCondition(dcName, "Updating", string(corev1.ConditionFalse))
+			ns.WaitForDatacenterOperatorProgress(dcName, "Ready", 1800)
+		})
+	})
+})
```
jimdickinson, line 69: our test framework is in great shape when we can do this test in under 100 lines! cc @sandoichi

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1609,6 +1682,85 @@ func (rc *ReconciliationContext) CheckRollingRestart() result.ReconcileResult {
...
+			logger.Error(err, "error patching datacenter status")
+			return result.Error(err)
+		}
+
+		// There may have been changes to the CassandraDatacenter resource that we ignored
+		// while executing some action on the cluster. For example, a user may have 
+		// requested to scale up the node count while we were in the middle of a rolling
+		// restart. To account for this, we requeue to ensure reconcile gets called again
+		// to pick up any such changes that we ignored previously.
+		return result.RequeueSoon(0)
```
jimdickinson, line 1757: are we already "ignoring" stuff?

johntrimble: Err not yet. Once we start doing things like postponing a scale up until a rolling restart is done, that line will become important. However, I ended up not doing the other half of this where the operator consumes the conditions to inform what it does and when.

Would you prefer I remove it?

jimdickinson: nah, leave it there, I just wanted to know

PR #59, opened 2020-04-24
KO-219 KO-360 KO-361 Handle adding racks

sandoichi:


operator/pkg/apis/cassandra/v1beta1/webhook.go
```diff
@@ -6,26 +6,36 @@ package v1beta1
...
 	"reflect"
+
+	"k8s.io/apimachinery/pkg/runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
 var log = logf.Log.WithName("api")
 
+func attemptedTo(action string, actionStrArgs ...interface{}) error {
```
sandoichi, line 18: Not sure if you guys like this type of abstraction or not.

johntrimble: :+1:

operator/pkg/apis/cassandra/v1beta1/webhook.go
```diff
@@ -65,18 +75,33 @@ func ValidateDatacenterFieldChanges(oldDc CassandraDatacenter, newDc CassandraDa
...
+
+		if newSizeDifference <= 0 {
+			return attemptedTo("add Rack without increasing Size")
+		}
+
+		if int(newSizeDifference) < newRackCount {
+			return attemptedTo(
+				fmt.Sprintf("add Racks without increasing Size enough to populate each rack.\n" +
+					"New racks added: %d, Size increased by: %d", newRackCount, newSizeDifference))
+		}
```
johntrimble, line 93: The line `newRackCount := len(newRacks) - len(oldRacks)` doesn't look right with the check `if int(newSizeDifference) < newRackCount`. How would this ever work if I added a single rack? Ideally, we'd leverage the same logic used in reconciliation to determine rack sizes from node counts so that validation stays aligned with reconciliation.

tests/delete_node_terminated_container/delete_node_terminate_container_suite_test.go
```diff
@@ -64,9 +64,9 @@ var _ = Describe(testName, func() {
 			ns.WaitForOutputAndLog(step, k, "Starting", 1200)
 
 			// give the cassandra container some time to get created
-			time.Sleep(20 * time.Second)
+			time.Sleep(10 * time.Second)
```
sandoichi, line 67: This test seemed to be a bit flaky and failed several times. I simplified the dc in this case to just be 1 node that the test is manipulating. During the failures, the node regained readiness and recovered from us messing with the mgmt api process while it was starting. I'm not sure if it was just the 20 seconds being too generous and the pod fully started in that time... but i figured reducing to single rack/single node and cutting the time buffer in half would give us less moving parts to investigate if it happens again.

operator/pkg/apis/cassandra/v1beta1/webhook.go
johntrimble: Err... wait, I misread this... still not sure if the logic is right here though. If I go from 9 nodes on 2 racks to 11 nodes on 3 racks, won't I get the following:

```
rack 1 -> 5 nodes
rack 2 -> 4 nodes
```

to

```
rack 1 -> 4 nodes
rack 2 -> 4 nodes
rack 3 -> 3 nodes
```

Causing rack 1 to have fewer nodes?

jimdickinson: yeah after thinking about this for too many hours, I think a simple way to phrase the solution is "you have to add as many nodes as the smallest rack had before the increase *times* the number of racks you're adding"

jimdickinson: My horrible ASCII art logic

```
9 nodes in 2 racks
xxxxx
xxxx

13 nodes in 3 racks
xxxxx
xxxx
xxxx
```

operator/pkg/apis/cassandra/v1beta1/webhook.go
jimdickinson: I like this, but should it be `actionStrArgs ...string` ?

sandoichi: nope, because then it wouldn't support `%d` and others. I modeled it after the `fmt.Sprintf` definition

jimdickinson: but you always do %s below?

jimdickinson: oh nvm, I misread

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -445,3 +445,19 @@ func (dc *CassandraDatacenter) GetContainerPorts() ([]corev1.ContainerPort, erro
 
 	return ports, nil
 }
+
+func SplitRacks(nodeCount, rackCount int) []int {
```
jimdickinson, line 449: 👍

operator/pkg/apis/cassandra/v1beta1/webhook.go
```diff
@@ -6,26 +6,36 @@ package v1beta1
...
 
+func attemptedTo(action string, actionStrArgs ...interface{}) error {
+	var msg string
+	if actionStrArgs != nil {
+		msg = fmt.Sprintf(action, actionStrArgs...)
+
+	} else {
+		msg = action
+	}
+	return fmt.Errorf("CassandraDatacenter attempted to %s", msg)
```
jimdickinson, line 26:
```suggestion
	return fmt.Errorf("CassandraDatacenter write rejected, attempted to %s", msg)
```

what do you think?

operator/pkg/apis/cassandra/v1beta1/webhook.go
```diff
@@ -35,24 +45,24 @@ func ValidateSingleDatacenter(dc CassandraDatacenter) error {
 func ValidateDatacenterFieldChanges(oldDc CassandraDatacenter, newDc CassandraDatacenter) error {
 
 	if oldDc.Spec.ClusterName != newDc.Spec.ClusterName {
-		return errors.New("CassandraDatacenter attempted to change ClusterName")
+		return attemptedTo("change ClusterName")
```
jimdickinson, line 48: getting nitpicky, do you think this should match the way users see the yaml? so `clusterName`, if so. also all the ones below here, if you agree.

operator/pkg/apis/cassandra/v1beta1/webhook.go
```diff
@@ -65,18 +75,38 @@ func ValidateDatacenterFieldChanges(oldDc CassandraDatacenter, newDc CassandraDa
...
-		return fmt.Errorf("CassandraDatacenter attempted to remove Rack")
+		return attemptedTo("remove Rack")
+	}
+
+	newRackCount := len(newRacks) - len(oldRacks)
+	if newRackCount > 0 {
+		newSizeDifference := newDc.Spec.Size - oldDc.Spec.Size
+		oldRackNodeSplit := SplitRacks(int(oldDc.Spec.Size), len(oldRacks))
+		minNodesFromOldRacks := oldRackNodeSplit[len(oldRackNodeSplit)-1]
+		minSizeAdjustment := minNodesFromOldRacks * newRackCount
```
jimdickinson, line 86: lgtm - sorry for the headache caused by the API

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -75,23 +75,17 @@ func (rc *ReconciliationContext) CalculateRackInformation() error {
 		return fmt.Errorf("assertion failed! rackCount should not possibly be zero here")
 	}
 
-	// nodes_per_rack = total_size / rack_count + 1 if rack_index < remainder
-
-	nodesPerRack, extraNodes := nodeCount/rackCount, nodeCount%rackCount
 	seedsPerRack, extraSeeds := seedCount/rackCount, seedCount%rackCount
```
jimdickinson, line 78: you could use

```suggestion
	rackSeedCounts := api.SplitRacks(seedCount, rackCount)
```

with some more changes below here too

tests/delete_node_terminated_container/delete_node_terminate_container_suite_test.go
jimdickinson: only this test?

operator/pkg/apis/cassandra/v1beta1/webhook.go
jimdickinson: yeah this is a great way to keep the messaging uniform!

tests/delete_node_terminated_container/delete_node_terminate_container_suite_test.go
sandoichi: yeah because this one relies on getting a node stuck in a `starting` state, which means it has to kill the mgmt api process in a limited time window in order for it to not be able to recover.

operator/pkg/reconciliation/reconcile_racks.go
sandoichi: ah good idea

operator/pkg/apis/cassandra/v1beta1/webhook.go
sandoichi: sounds good

tests/delete_node_terminated_container/delete_node_terminate_container_suite_test.go
jimdickinson: 👍

PR #60, opened 2020-04-27
KO-396 remove operator init

sandoichi:
Ripped out old init container code.


mage/kind/lib.go
```diff
@@ -83,6 +82,15 @@ func loadSettings() {
 	}
 }
 
+// There is potential for globally scoped resources to be left
+// over from a previous helm install
+func cleanupLingeringHelmResources() {
```
jimdickinson, line 87: 💯

PR #65, opened 2020-04-28
KO-398 push to GH packages instead of artifactory

sandoichi:


sandoichi:
> Looks good to me. Have we tested it?

Yeah you can see the list of images published via github actions here: https://github.com/datastax/cass-operator/packages/174279/versions

Issue #69, opened 2020-05-01
Some creation events missing after creating a cluster


jsanda:
I am using v1.1.0 of the operator. I created a cluster in GKE. Here is my cassdc spec:

```
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: multi-rack
spec:
  clusterName: multi-rack
  serverType: cassandra
  serverVersion: 3.11.6
  managementApiAuth:
    insecure: {}
  size: 9
  racks:
  - name: us-east1-b
    zone: us-east1-b
  - name: us-east1-c
    zone: us-east1-c
  - name: us-east1-d
    zone: us-east1-d
  storageConfig:
    cassandraDataVolumeClaimSpec:
      storageClassName: server
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
```

I was monitor the provisioning progress with `kubectl -n cass-operator get -w cassdc multi-rack -o yaml`.

Once I saw `.status.cassandraOperatorProgress: Ready` I checked the events and here is what I saw,

```
$ kubectl -n cass-operator describe cassdc multi-rack
Name:         multi-rack
Namespace:    cass-operator
Labels:       <none>
Annotations:  kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"cassandra.datastax.com/v1beta1","kind":"CassandraDatacenter","metadata":{"annotations":{},"name":"multi-rack","namespace":"...
API Version:  cassandra.datastax.com/v1beta1
Kind:         CassandraDatacenter
Metadata:
  Creation Timestamp:  2020-05-01T16:28:04Z
  Finalizers:
    finalizer.cassandra.datastax.com
  Generation:        2
  Resource Version:  21054
  Self Link:         /apis/cassandra.datastax.com/v1beta1/namespaces/cass-operator/cassandradatacenters/multi-rack
  UID:               1af119f8-af42-4689-afca-dfcaf601ced0
Spec:
  Cluster Name:  multi-rack
  Management API Auth:
    Insecure:
  Racks:
    Name:  us-east1-b
    Zone:  us-east1-b
    Name:  us-east1-c
    Zone:  us-east1-c
    Name:  us-east1-d
    Zone:  us-east1-d
  Resources:
  Server Type:     cassandra
  Server Version:  3.11.6
  Size:            9
  Storage Config:
    Cassandra Data Volume Claim Spec:
      Access Modes:
        ReadWriteOnce
[...]

PR #76, opened 2020-05-05
Add logging of all events for more reliable retrieval

johntrimble:
Addresses #69 by logging any events that are produced to ensure they can be retrieved even if dropped due to an excessive number being produced.


operator/pkg/events/events_test.go
```diff
@@ -0,0 +1,69 @@
...
+	values := append([]interface{}{msg}, keysAndValues...)
+	_ = l.Called(values...)
+}
+
+func (l *MockInfoLogger) Enabled() bool {
+	args := l.Called()
+  	return args.Bool(0)
+}
+
+func TestLoggingEventRecorder(t *testing.T) {
```
jimdickinson, line 27: great use of the mocking lib

Issue #72, opened 2020-05-01
CassandraDatacenter status updates not happening correctly during cluster creation


jsanda:
I have reproduced this behavior consistently 3 or 4 times on a GKE cluster with 9 nodes spread across 3 zones. I am using v1.1.0 of the operator. Here is my cassdc spec:

```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: multi-rack
spec:
  clusterName: multi-rack
  serverType: cassandra
  serverVersion: 3.11.6
  managementApiAuth:
    insecure: {}
  size: 9
  racks:
  - name: us-east1-b
    zone: us-east1-b
  - name: us-east1-c
    zone: us-east1-c
  - name: us-east1-d
    zone: us-east1-d
  storageConfig:
    cassandraDataVolumeClaimSpec:
      storageClassName: server
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
```

I was watching status updates in the console and I have seen multiple times now. I will show the last few status updates to give a better picture of what is happening.

With this first update, we see that 8 pods are reported:

```yaml
status:
  cassandraOperatorProgress: Updating
  lastRollingRestart: "2020-05-01T21:24:53Z"
  lastServerNodeStarted: "2020-05-01T21:33:08Z"
  nodeStatuses:
    multi-rack-multi-rack-us-east1-b-sts-0:
      hostID: 42a10825-ffca-43e0-8a44-13440fc0a586
      nodeIP: 10.28.8.4
    multi-rack-multi-rack-us-east1-b-sts-1:
      hostID: 6e496a9b-8a8a-43f7-9149-83ae15633cc1
      nodeIP: 10.28.3.3
    multi-rack-multi-rack-us-east1-b-sts-2:
      hostID: f3008646-ea00-4d74-9575-c90647a0811e
      nodeIP: 10.28.5.4
    multi-rack-multi-rack-us-east1-c-sts-0:
      hostID: d664763f-0e4c-4511-a720-14a494742de7
      nodeIP: 10.28.6.4
    multi-rack-multi-rack-us-east1-c-sts-2:
      hostID: f73691e4-b0b3-47fb-89c2-e7e6ecc62901
      nodeIP: 10.28.4.10
    multi-rack-multi-rack-us-east1-d-sts-0:
      hostID: 3cab2810-be05-497a-b22d-a98f286e181b
      nodeIP: 10.28.1.4
[...]

PR #77, opened 2020-05-06
requeue after the last node has its node-state label set to Started during cluster creation

jsanda:
This PR is for #72. It fixes an issue in which we basically see out of order status updates. Specifically, the `.Status.SuperUserUpserted` property is getting set before the last node is added to `.Status.NodeStatuses`. The actual order of operations is correct though. The super user is not created until all C* nodes are started. It is just that the status updates basically occur out of order.


tests/scale_up_status_updates/scale_up_status_updates_suite.go
```diff
@@ -0,0 +1,60 @@
...
+			By("creating a namespace")
+			err := kubectl.CreateNamespace(namespace).ExecV()
+			Expect(err).ToNot(HaveOccurred())
+
+			step := "setting up cass-operator resources via helm chart"
+			ns.HelmInstall("../../charts/cass-operator-chart")
+
+			ns.WaitForOperatorReady()
+
+			step = "creating a datacenter resource with 2 racks/3 nodes"
```
sandoichi: step should say `2 racks/6 nodes` right?

tests/scale_up_status_updates/scale_up_status_updates_suite_test.go
```diff
@@ -0,0 +1,60 @@
...
+			By("creating a namespace")
+			err := kubectl.CreateNamespace(namespace).ExecV()
+			Expect(err).ToNot(HaveOccurred())
+
+			step := "setting up cass-operator resources via helm chart"
+			ns.HelmInstall("../../charts/cass-operator-chart")
+
+			ns.WaitForOperatorReady()
+
+			step = "creating a datacenter resource with 2 racks/3 nodes"
```
sandoichi, line 49: step should say 2 racks/6 nodes right?

tests/scale_up_status_updates/scale_up_status_updates_suite_test.go
```diff
@@ -0,0 +1,60 @@
...
+		ns.Terminate()
+	})
+
+	RegisterFailHandler(Fail)
+	RunSpecs(t, testName)
+}
+
+var _ = Describe(testName, func() {
+	Context("when in a new cluster", func() {
+		Specify("the operator can scale up a datacenter", func() {
```
sandoichi, line 39: Might want to make this a bit more descriptive for what you are actually testing. We have a couple of other scale up tests that do nothing but increase the node count, and we probably want to distinguish the messages in this one.

tests/scale_up_status_updates/scale_up_status_updates_suite_test.go
```diff
@@ -0,0 +1,60 @@
...
+			step = "creating a datacenter resource with 2 racks/3 nodes"
+			k := kubectl.ApplyFiles(dcYaml)
+			ns.ExecAndLog(step, k)
+
+			ns.WaitForSuperUserUpserted(dcName, 600)
+
+			step = "checking that all nodes have been started"
+			nodeStatusesHostIds := ns.GetNodeStatusesHostIds(dcName)
+			Expect(len(nodeStatusesHostIds), 6)
+		})
```
sandoichi, line 58: Most of our tests will typically test that the dc can be successfully deleted at the end of the test scenario. It probably isn't super important to have on every single one (especially since you really aren't doing any crazy patching or unusual things) so I'll let you decide if you want to add the logic to this test or not. See [our scale up test](https://github.com/datastax/cass-operator/blob/master/tests/scale_up/scale_up_suite_test.go#L90-L99) for reference if you choose to (again, purely optional from my perspective).

tests/scale_up_status_updates/scale_up_status_updates_suite_test.go
jsanda: fixed

tests/scale_up_status_updates/scale_up_status_updates_suite_test.go
jsanda: yeah, I just copied/pasted from the scale up test. I will update this.

tests/scale_up_status_updates/scale_up_status_updates_suite_test.go
jsanda: I will add the deletion check(s). I would rather be consistent with other tests.

tests/scale_up_status_updates/scale_up_status_updates_suite_test.go
jsanda: @sandoichi wdyt about the updated message?

sandoichi: Looks good

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -476,6 +476,9 @@ func (rc *ReconciliationContext) CheckPodsReady(endpointData httphelper.CassMeta
 	desiredSize := int(rc.Datacenter.Spec.Size)
 
 	if desiredSize == readyPodCount && desiredSize == startedLabelCount {
+		if nodeStarted {
```
jimdickinson, line 479: I think you should add a one line comment here

also why not move this directly under
```
	nodeIsStarting, nodeStarted, err := rc.findStartingNodes()
```
?

jsanda: comment added.

I hadn't really considered moving the `if` block. That will require some further discussion to ensure I don't wind up introducing some regressions.

jimdickinson: yeah let's not introduce instability, I can tackle this when refactoring some stuff in an upcoming PR

jsanda:
I ran all integration tests locally. I get one failure:

```
• Failure [1244.484 seconds]
Add racks
/Users/jsanda/Development/gocode/src/github.com/datastax/cass-operator/tests/add_racks/add_racks_suite_test.go:40
  when in a new cluster
  /Users/jsanda/Development/gocode/src/github.com/datastax/cass-operator/tests/add_racks/add_racks_suite_test.go:41
    racks can be added if the size is increased accordingly [It]
    /Users/jsanda/Development/gocode/src/github.com/datastax/cass-operator/tests/add_racks/add_racks_suite_test.go:42

    Unexpected error:
        <*errors.errorString | 0xc000246390>: {
            s: "Timed out waiting for value. Expected to output to contain: true, but got .",
        }
        Timed out waiting for value. Expected to output to contain: true, but got .
    occurred
```
I also ran the tests in master and got the same failure.

jimdickinson:
We'll see if thats a GKE vs KIND issue...

PR #79, opened 2020-05-07
Support k3d + abstract common targets from all cluster types

sandoichi:
The goal of this is to make a simple set of k8s mage targets that can be used across any cluster type (kind, k3d, gcp) and you just pick your cluster type by setting an env var.

This reduces duplicate logic for each cluster type, and drastically cuts down on how many mage targets we have.

TODO: still need to write docs


mage/k8s/lib.go
```diff
@@ -0,0 +1,260 @@
+// Copyright DataStax, Inc.
```
sandoichi, line 1: A large portion of the code in this file has been directly ripped out of `./mage/kind/lib.go`

mage/k8s/lib.go
```diff
@@ -0,0 +1,260 @@
...
+//
+// Set M_LOAD_DEV_IMAGES to "true" to pull and
+// load the dev images listed in buildsettings.yaml
+// into the cluster.
+func SetupEmptyCluster() {
+	loadClusterSettings()
+	clusterActions.DeleteCluster()
+	clusterActions.CreateCluster()
+	// some clusters need a few seconds before we can get kubeconfig info
+	time.Sleep(5 * time.Second)
```
respringer, line 93: On my laptop, I had to change this timeout to 10.

mage/k8s/lib.go
```diff
@@ -0,0 +1,260 @@
...
+// This must either be applied manually or by calling SetupCassandraCluster
+// or SetupDSECluster.
+// This target assumes that helm is installed and available on path.
+func SetupExampleCluster() {
+	mg.Deps(SetupEmptyCluster)
+	kubectl.CreateSecretLiteral("cassandra-superuser-secret", "devuser", "devpass").ExecVPanic()
+
+	var namespace = "default"
+	var overrides = map[string]string{"image": "datastax/cass-operator:latest"}
+	err := helm_util.Install("./charts/cass-operator-chart", "cass-operator", namespace, overrides)
```
respringer, line 136: Just an FYI: I had to change this to a function to invoke a Makefile "deploy" target for astra operator.

docs/developer/k8s_targets.md
```diff
@@ -0,0 +1,129 @@
...
+
+After your k8s cluster has been created, you can configure kubectl to point to it.
+
+### Bash shell
+```bash
+M_K8S_FLAVOR=k3d mage k8s:kubeconfig
+```
+
+Note: this target currently assumes that you are using bash. Different shells may
+require manual configuration for the kubeconfig
```
johntrimble, line 129: Are the docs for running integration tests not impacted by this? Do they still just use whatever cluster the KUBECONFIG env var points to?

mage/config/lib.go
```diff
@@ -65,3 +65,39 @@ func ReadBuildSettings() BuildSettings {
...
+	DeleteCluster       func() error
+	ClusterExists       func() bool
+	CreateCluster       func()
+	LoadImage           func(image string)
+	InstallTool         func()
+	ReloadLocalImage    func(image string)
+	ApplyDefaultStorage func()
+	SetupKubeconfig     func()
+	DescribeEnv         func() map[string]string
+}
```
johntrimble, line 79: :100:

docs/developer/k8s_targets.md
sandoichi: ah good catch, there are some of the "set up the world for me in KIND" targets listed in the integ docs that i gotta update. Regular `mage integ:...` will remain unchanged though

README.md
```diff
@@ -242,11 +242,13 @@ cluster with six or more worker nodes.
...
 There are a number of ways to run the operator, see the following docs for
 more information:
-* [kind](docs/developer/kind.md): Kubernetes in Docker is the recommended
-  Kubernetes distribution for use by software engineers working on the operator.
-  KIND can simulate a k8s cluster with multiple worker nodes on a single
-  physical machine, though it's necessary to dial down the database memory
-  requests.
+* [k8s targets](docs/developer/k8s_targets.md): A set of mage targets for
+  automating a variety of tasks for several different supported k8s flavors.
+  At the moment, we support KIND, k3d, and gke. These targets can setup and
```
jimdickinson, line 247: I love the good old rule of three we're hitting for the interface here :)

docs/developer/k8s_targets.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+A supported flavor can then be used by setting the `M_K8S_FLAVOR` env variable to that
+flavor, for example:
+
+```bash
+M_K8S_FLAVOR=kind
+```
+
+If the `M_K8S_FLAVOR` is not specified, the default flavor becomes `k3d`
```
jimdickinson, line 24: 😮 👍

mage/config/lib.go
jimdickinson: 🥇

PR #80, opened 2020-05-08
KO-382 cleanup after scaleup

sandoichi:


operator/pkg/httphelper/client.go
```diff
@@ -159,7 +163,42 @@ func (client *NodeMgmtClient) CallDrainEndpoint(pod *corev1.Pod) error {
...
+
+	request := nodeMgmtRequest{
+		endpoint: fmt.Sprintf("/api/v0/ops/keyspace/cleanup"),
+		host:     BuildPodHostFromPod(pod),
+		method:   http.MethodPost,
+		timeout:  time.Second * 20,
+		body:     body,
+	}
+
+	_, err = callNodeMgmtEndpoint(client, request, "application/json")
```
jimdickinson, line 201: there is an `http` constant for this, I think

operator/pkg/httphelper/client.go
```diff
@@ -159,7 +163,42 @@ func (client *NodeMgmtClient) CallDrainEndpoint(pod *corev1.Pod) error {
...
+		postData["jobs"] = strconv.Itoa(jobs)
+	}
+
+	if keyspaceName != "" {
+		postData["keyspace_name"] = keyspaceName
+	}
+
+	if len(tables) > 0 {
+		postData["tables"] = tables
+	}
```
jimdickinson, line 186: Do you think this is nicer? It's not exactly the same logic, yet
```
postData := struct {
    Jobs string `json:"jobs"`
    Keyspace string `json:"keyspace_name"`
    Tables []string `json:"tables"`
}{
    Jobs: strconv.Itoa(jobs),
    Keyspace: keyspaceName,
    Tables: tables,
}
```

PR #87, opened 2020-05-13
Merge k3d config with default/current kubeconfig

sandoichi:
When using k3d, it creates it own separate config file instead of writing to your current one. This change will merge your current (or default) kubeconfig with the k3d one

### DIFF
--- a/mage/k3d/lib.go
+++ b/mage/k3d/lib.go
@@ -5,6 +5,7 @@ package k3d
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


mage/k3d/lib.go
```diff
@@ -103,9 +104,28 @@ func reloadLocalImage(image string) {
...
+
+	// merge current config + new k3d config
+	// the new config must appear before the current config to update it
+	// when merging
+	os.Setenv("KUBECONFIG", fmt.Sprintf("%s:%s", k3dConfig, currentConfig))
+	merged := shutil.OutputPanic("kubectl", "config", "view", "--flatten")
+
+	// look up current config file so we can retain its permissions
+	// before overwriting
+	configInfo, err := os.Stat(currentConfig)
```
sandoichi, line 118: this could technically error out if your kubeconfig wasn't specified and `~/.kube/` dir did not contain a `config` file... but i'm not sure if its worth adding logic to create an empty default in that case (is that even a real concern?)

mage/k3d/lib.go
```diff
@@ -103,9 +104,28 @@ func reloadLocalImage(image string) {
...
+	mageutil.PanicOnError(err)
+
+	// overwrite current config file with the merged file
+	err = ioutil.WriteFile(currentConfig, []byte(fmt.Sprintf("%s\n", merged)), configInfo.Mode())
+	if err != nil {
+		fmt.Printf("Failed to write kubeconfig file at %s\n", currentConfig)
+		panic(err)
+	}
+
+	kubectl.ClusterInfoForContext("default").ExecVPanic()
```
sandoichi, line 128: k3d calls their cluster context `default` in its config, even if you specify custom cluster names, the context is still `default`. Hopefully this doesn't conflict with any other flavors in the future...

jimdickinson: renaming it is just hacking on some yaml, fwiw

mage/k3d/lib.go
jimdickinson: I'm sure someone will hit that, we can just document somewhere they should `mkdir -p ~/.kube && touch ~/.kube/config`

mage/k3d/lib.go
```diff
@@ -103,9 +104,28 @@ func reloadLocalImage(image string) {
...
-	os.Setenv("KUBECONFIG", config)
-	fmt.Printf("To set up kubectl in your shell to use this cluster, set:\n\tKUBECONFIG=%s\n", config)
+	k3dConfig := shutil.OutputPanic("k3d", "get-kubeconfig")
+	currentConfig := kubectl.GetKubeconfig()
+
+	// merge current config + new k3d config
+	// the new config must appear before the current config to update it
+	// when merging
+	os.Setenv("KUBECONFIG", fmt.Sprintf("%s:%s", k3dConfig, currentConfig))
+	merged := shutil.OutputPanic("kubectl", "config", "view", "--flatten")
```
johntrimble, line 114: I don't quite follow this. This line suggests that `KUBECONFIG` can have multiple values, so why do we need to merge the kube config files at all?

sandoichi: So the goal of this is to set up the user's kubeconfig so that they can immediately start using kubectl to talk to the new cluster that we just stood up for them. We can easily have kubeconfig set up inside our go process by doing the normal os.Setenv call, but as soon as the go process finished running, the user's shell would not be set up to let them start using kubectl without running at least 1 manual command to set up kubeconfig.

By doing it this way, we are leveraging kubectl's ability to merge multiple configs and then we update the user's current config on disk with the merged value, and then finally we tell kubectl to switch to the new context. 

KIND and gcloud already did this automatically when you create a new cluster, but k3d was more conservative and created a separate config file and didn't do any merging.

PR #91, opened 2020-05-15
Create default kubeconfig when setting up k3d

sandoichi:
If kubeconfig isn't specified, and default config doesn't exist, create it

### DIFF
--- a/mage/kubectl/lib.go
+++ b/mage/kubectl/lib.go
@@ -5,13 +5,13 @@ package kubectl
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"os/user"
 	"regexp"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mage/kubectl/lib.go
```diff
@@ -14,13 +14,23 @@ import (
 	mageutil "github.com/datastax/cass-operator/mage/util"
 )
 
-func GetKubeconfig() string {
+func GetKubeconfig(createDefault bool) string {
```
sandoichi, line 17: not sure if jamming the logic into this function is the best idea or not, but it does save from having to lookup user dir in multiple places

jimdickinson: yeah I think this is fine

PR #94, opened 2020-05-19
Add ability to specify additional users

johntrimble:


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -212,12 +220,17 @@ func NewDatacenterCondition(conditionType DatacenterConditionType, status corev1
 	}
 }
 
+// Note: The below property 'SuperUserUpserted' needs to be refactored 
+// at some point to treat superuser as one word instead of two. Also, 
+// this is not just for the one superuser any more, but all users
+// defined on the CassandraDatacenter resource.
```
johntrimble, line 226: I didn't want this comment dumped in the CRD, so I put it up here instead.

mage/ginkgo/lib.go
```diff
@@ -21,7 +21,8 @@ import (
 )
 
 const (
-	EnvNoCleanup = "M_NO_CLEANUP"
+	EnvNoCleanup  = "M_NO_CLEANUP"
+	OperatorImage = "datastax/cass-operator:latest"
```
johntrimble, line 25: `sed` friendly for testing with Kubespray + OpenStack

mage/helm/lib.go
```diff
@@ -12,6 +12,9 @@ import (
 func Install(chartPath string, releaseName string, namespace string, overrides map[string]string) error {
 	args := []string{
 		"install",
+		// k8s 1.15 does not support all the validations we have setup on 
+		// the CRD, so we just turn off validation. What could go wrong?
+		"--disable-openapi-validation",
```
johntrimble, line 17: If only the old versions of things died and were gone forever.

operator/deploy/crds/cassandra.datastax.com_cassandradatacenters_crd.yaml
```diff
@@ -6123,8 +6136,8 @@ spec:
                 type: object
               type: object
             superUserUpserted:
-              description: The timestamp at which CQL superuser credentials were last
-                upserted to the management API
+              description: The timestamp at which cassandra user credentials were
+                last upserted to the management API
```
johntrimble, line 6140: This is overloaded now. We use the same timestamp for both updating the standard superuser as well as any additional users specified on the CRD. I've not changed the field name for backwards compatibility. We could add a separate field for updates to the other users, but that seems unnecessary.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -67,6 +67,11 @@ func getImageForServerVersion(server, version string) (string, error) {
 	return "", err
 }
 
+type CassandraUser struct {
+	SecretName string `json:"secretName"`
+	Superuser bool `json:"superuser"`
+}
```
johntrimble, line 73: We are assuming that all users specified here should have login enabled

operator/pkg/reconciliation/handler.go
```diff
@@ -159,8 +159,12 @@ func (rc *ReconciliationContext) isValid(dc *api.CassandraDatacenter) error {
 
 	// Basic validation up here
 
+	// validate the required superuser
 	errs = append(errs, rc.validateSuperuserSecret()...)
 
+	// validate any other defined users
+	errs = append(errs, rc.validateCassandraUserSecrets()...)
```
johntrimble, line 166: I was thinking this might not be super fantastic for performance reasons, but then we load every pod resource for the `CassandraDatacenter` on every reconcile loop, so I guess it's fine.

operator/pkg/reconciliation/secrets_test.go
```diff
@@ -84,17 +84,6 @@ func Test_validateSuperuserSecretContent(t *testing.T) {
...
-			secretNil: true,
-			valid:     true,
-			message:   "validation should pass when a secret is going to be auto-generated",
-		},
-		{
-			superuserSecret: "my-fun-secret",
-			secretNil:       true,
-			valid:           false,
-			message:         "validation should fail when the user explicitly specifies a secret but it does not exist",
-		},
```
johntrimble, line 97: Checks for these were moved outside `validateSuperuserSecretContent()` so removing these cases here.

mage/helm/lib.go
jimdickinson: 😬 I don't feel good about this - what happens if you remove it?

operator/deploy/crds/cassandra.datastax.com_cassandradatacenters_crd.yaml
jimdickinson: I think you can reset the status bits for the user creation. Make whatever is ideal for you, and label the existing stuff deprecated.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
jimdickinson: I think you can reset the status bits for the user creation. Make whatever is ideal for you, and label the existing stuff deprecated.

mage/helm/lib.go
johntrimble: It will fail to install in 1.15.

jimdickinson: ok - just add a FIXME, I'm working on better handling multiple k8s versions in our helm chart right now, so I'll come back here when that's ready

operator/pkg/reconciliation/handler.go
jimdickinson: good thought - let's make a ticket to stress test this a little with 50 users

tests/superuser-secret-provided/superuser_secret_provided_test.go
```diff
@@ -21,6 +21,8 @@ var (
 	defaultSecretName   = "cluster2-superuser"
 	superuserName       = "bob"
 	superuserPass       = "bobber"
+	bobbyuserName       = "bobby"
+	bobbyuserPass       = "littlebobbydroptables"
```
jimdickinson, line 25: 💯 🤣 🙌

operator/pkg/reconciliation/handler.go
johntrimble: KO-431

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -222,11 +230,17 @@ func NewDatacenterCondition(conditionType DatacenterConditionType, status corev1
...
 
-	// The timestamp at which CQL superuser credentials
-	// were last upserted to the management API
+	// Deprecated. Use usersUpserted instead. The timestamp at 
+	// which CQL superuser credentials were last upserted to the 
+	// management API
 	// +optional
 	SuperUserUpserted metav1.Time `json:"superUserUpserted,omitempty"`
 
+	// The timestamp at which managed cassandra users' crednetials
```
jimdickinson, line 239:
```suggestion
	// The timestamp at which managed cassandra users' credentials
```

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -676,52 +676,84 @@ func (rc *ReconciliationContext) CheckRackPodLabels() result.ReconcileResult {
...
+
+	if dc.Spec.Stopped {
+		rc.ReqLogger.Info("cluster is stopped, skipping CreateUser")
+		return result.Continue()
+	}
+
+	//Skip upsert if already did so recently
+	if !shouldUpsertUsers(*dc) {
+		rc.ReqLogger.Info(fmt.Sprintf("The users were last upserted at %v, skipping upsert", 
+			dc.Status.UsersUpserted))
```
jimdickinson, line 722: Better to keep the logging structured...

```suggestion
		rc.ReqLogger.Info("users upserted recently, skipping upsert", 
			"lastUpsert", dc.Status.UsersUpserted)
```

PR #95, opened 2020-05-21
support universal base image

sandoichi:
The bulk of the changes boils down to:
Specify `MO_BASE_OS` when running the mage operator build targets or the mage k8s cluster setup targets and it will build an operator based off of the image you specify in `MO_BASE_OS`, tag it with a `-ubi` suffix, and then properly load it into your local cluster + override the helm chart value to use that image in the deployment.

####  example to build the operator and get `-ubi` tagged images:
```
MO_BASE_OS=registry.access.redhat.com/ubi7/ubi-minimal:7.8 mage operator:build
```

####  example to build the operator, setup a k3d cluster, and load the image in + stand up the deployment:
```
MO_BASE_OS=registry.access.redhat.com/ubi7/ubi-minimal:7.8 mage k8s:setupExampleCluster
```

The operator built for the ubi image will then always choose ubi based dse/config builder/logging container images when it stands up a datacenter.

We will now also always publish both regular and `-ubi` versions of the operator from the github workflow.


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -46,25 +47,51 @@ const (
...
-		return cassandra_4_0_0, nil
+		return cassandra_4_0_0, true
 	}
-	err := fmt.Errorf("server '%s' and version '%s' do not work together", server, version)
-	return "", err
+	return "", false
+}
+
+func getImageForUniversalBaseOs(sv string) (string, bool) {
+	panic("TODO: Need universal basic image coords")
```
sandoichi, line 94: Did we decide that we will want to point at dockerhub hosted images here? or redhat's own registry?

jimdickinson: I prefer docker hub

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -46,25 +47,51 @@ const (
...
-		dse_6_8_0        = "datastax/dse-server:6.8.0"
-	)
-	sv := server + "-" + version
+	baseImageOs := os.Getenv(envBaseImageOs)
+
+	var imageCalc func(string) (string, bool)
+	var img string
+	var success bool
+
+	if baseImageOs == "" {
```
jimdickinson, line 66: but even alpine wouldn't be an empty string?

sandoichi: so we re introduce the arg in the last stage of the dockerfile with a default of `""` so alpine is the initial default, then the `from` will use it in stage 2, then it will go back to `""` if its not manually set.

This way our default is always `""` whether its alpine or anything else.

jimdickinson: this is pretty clever! 👍

.github/workflows/operatorPRBuildAndDeploy.yml
```diff
@@ -47,7 +47,34 @@ jobs:
...
+        if: github.event.pull_request.head.repo.full_name == 'datastax/cass-operator'
+        env:
+          MO_GH_USR: 'datastax/cass-operator'
+          MO_GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          MO_GH_PKG_REPO: 'datastax/cass-operator/operator'
+        run: |
+          export PATH=$GOROOT/bin:$GOPATH/bin:$PATH
+          export MO_TAGS=$(cat ./build/tagsToPush.txt)
+          mage operator:deployToGHPackages
+      - name: Build docker - ubi image
```
sandoichi, line 60: kind of meh to have these 3 steps duplicated for the UBI stuff. The alternative would be to add an extra env var/logic to the mage lib to support "build and tag both in one command". I could go either way on that.... but it would just be more conditionals and env stuff to keep track of

jimdickinson: Actions does have a way to loop over things if that would be neater

operator/pkg/reconciliation/constructor.go
```diff
@@ -404,7 +405,11 @@ func buildContainers(dc *api.CassandraDatacenter, serverVolumeMounts []corev1.Vo
 	// server logger container
 	loggerContainer := corev1.Container{}
 	loggerContainer.Name = "server-system-logger"
-	loggerContainer.Image = "busybox"
+	if baseImageOs := os.Getenv(api.EnvBaseImageOs); baseImageOs != "" {
+		loggerContainer.Image = baseImageOs
+	} else {
+		loggerContainer.Image = "busybox"
```
jimdickinson, line 411: not for this PR, but we should offer a way to shut off this tailing container

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
jimdickinson: for posterity - we decided that pointing to the RedHat repo was the right thing to do

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -46,25 +45,71 @@ const (
...
 type ProgressState string
 
+const (
+	defaultConfigBuilderImage     = "datastax/cass-config-builder:1.0.0"
+	cassandra_3_11_6              = "datastax/cassandra-mgmtapi-3_11_6:v0.1.2"
+	cassandra_4_0_0               = "datastax/cassandra-mgmtapi-4_0_0:v0.1.2"
+	dse_6_8_0                     = "datastax/dse-server:6.8.0"
+	ubi_cassandra_3_11_6          = "registry.connect.redhat.com/datastax/cassandra:3.11.6"
+	ubi_cassandra_4_0_0           = "TODO"
+	ubi_dse_6_8_0                 = "registry.connect.redhat.com/datastax/dse-server:6.8.0"
```
bradfordcp, line 55: 6.8.1 has been released, should this be updated or a separate PR?

bradfordcp: This approach doesn't seem very tenable as we move forward with more versions. I'm still of the opinion we should simply override the namespace or registry portion of the coordinates.

**Namespace:**
_datastax_/repo\_name:tag

Here _datastax_ is the namespace. It blindly assumes the repo names are the same.

**Repository**
_datastax/repo\_name_:tag

With this approach, the entire repository must be updated. It will require multiple constants, but it feels a little more organic as new versions of components come out.

My preference would be to store the repositories in constants and choose the appropriate value based on the UBI flag passed into the operator.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -46,25 +45,71 @@ const (
...
-		return dse_6_8_0, nil
+		return ubi_dse_6_8_0, true
 	case "cassandra-3.11.6":
-		return cassandra_3_11_6, nil
+		return ubi_cassandra_3_11_6, true
 	case "cassandra-4.0.0":
-		return cassandra_4_0_0, nil
+		//TODO no image available yet
+		return ubi_cassandra_4_0_0, false
 	}
```
bradfordcp, line 111: This still feels a little clunky. IE there are no changes in 6.8.1 that the operator cares about, but it feels like we can't run it with this approach.

operator/pkg/reconciliation/constructor.go
bradfordcp: Agreed

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
bradfordcp: or even just have environment variables available to override the default and ignore UBI entirely in the codebase.

jimdickinson: I'm gonna put 6.8.1 in another PR - don't want to do that until this one is merged

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -46,25 +45,71 @@ const (
...
 
+const (
+	defaultConfigBuilderImage     = "datastax/cass-config-builder:1.0.0"
+	cassandra_3_11_6              = "datastax/cassandra-mgmtapi-3_11_6:v0.1.2"
+	cassandra_4_0_0               = "datastax/cassandra-mgmtapi-4_0_0:v0.1.2"
+	dse_6_8_0                     = "datastax/dse-server:6.8.0"
+	ubi_cassandra_3_11_6          = "registry.connect.redhat.com/datastax/cassandra:3.11.6"
+	ubi_cassandra_4_0_0           = "TODO"
+	ubi_dse_6_8_0                 = "registry.connect.redhat.com/datastax/dse-server:6.8.0"
+	ubi_defaultConfigBuilderImage = "registry.connect.redhat.com/datastax/cass-config-builder:1.0.0"
```
jimdickinson, line 56: we're gonna stick with dockerhub, no need for `registry.connect.redhat.com` because it'll be hard to test - this PR is focusing on UBI bytes, not image hosting

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
bradfordcp: Creeper gonna creep

This was correctly identified as scope creep. Please ignore for now as private registries will go into a separate issue for consideration later.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
bradfordcp: More scope creepin'. Resolving

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
bradfordcp: Note the images on DockerHub are suffixed with `-ubi7`

.github/workflows/operatorStableBuildAndDeploy.yml
```diff
@@ -1,56 +0,0 @@
-name: Cass Operator Stable Build & Deploy
```
jimdickinson, line 1: 🙌  less duplication!

eldondevat:
readBaseOsIntoEnv

operator/cmd/manager/main.go
```diff
@@ -200,6 +206,36 @@ func main() {
...
+
+	rawVal, err := ioutil.ReadFile(baseOsArgFilePath)
+	if err != nil {
+		msg := fmt.Sprintf("Failed to read base OS arg file at %s", baseOsArgFilePath)
+		err = fmt.Errorf("%s. %v", msg, err)
+		return err
+	}
+
+	baseOs := strings.TrimSpace(string(rawVal))
+	os.Setenv("BASE_IMAGE_OS", baseOs)
```
eldondevat, line 233: Is there a reason not to use the `EnvBaseImageOs` constant here?

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -46,30 +45,75 @@ const (
...
+		return cassandra_3_11_6, true
+	case "cassandra-4.0.0":
+		return cassandra_4_0_0, true
+	}
+	return "", false
+}
+
+func getImageForUniversalBaseOs(sv string) (string, bool) {
 	switch sv {
 	case "dse-6.8.0":
```
eldondevat, line 104: What if we turn this into a map somewhere at the top of the file, so we don't have to dig all the way down into these functions to update it?

operator/cmd/manager/main.go
sandoichi: nope, good catch!

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
sandoichi: I think we have plans to update some of this for 6.8.1 support in a different PR.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -46,30 +45,75 @@ const (
 // This type exists so there's no chance of pushing random strings to our progress status
 type ProgressState string
 
+const (
+	defaultConfigBuilderImage     = "datastax/cass-config-builder:1.0.0"
+	cassandra_3_11_6              = "datastax/cassandra-mgmtapi-3_11_6:v0.1.5"
+	cassandra_4_0_0               = "datastax/cassandra-mgmtapi-4_0_0:v0.1.5"
+	dse_6_8_0                     = "datastax/dse-server:6.8.0"
+	ubi_cassandra_3_11_6          = "datastax/cassandra:3.11.6-ubi7"
+	ubi_cassandra_4_0_0           = "datastax/cassandra:4.0-ubi7"
```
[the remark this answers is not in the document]
sandoichi: Probably fine to leave as is for now to get this merged and then update it once we have the actual coordinate in place

PR #96, opened 2020-05-27
KO-430 remove NodeIP from status

johntrimble:
- Remove NodeIP from CassandraNodeStatus as it is not used (KO-430)
- Ensure datacenter not marked as ready until all host IDs recorded (fixes issues #72)
- Attempt to collect host ID information as soon as possible (KO-348)


operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -573,11 +573,6 @@ func (rc *ReconciliationContext) CheckPodsReady(endpointData httphelper.CassMeta
 	desiredSize := int(rc.Datacenter.Spec.Size)
 
 	if desiredSize == readyPodCount && desiredSize == startedLabelCount {
-		// When the ready and started counts match the desired counts and nodeStarted is true, then that means we have
-		// just started the last node in the data center.
-		if nodeStarted {
-			return result.RequeueSoon(2)
-		}
```
johntrimble, line 580: This is both probably not necessary due to earlier bug fixes and not sufficiently explicit. We are completely dependent on the comment here to understand what this is about. This has been replaced with `CheckCassandraNodeStatuses()` which makes explicit that we must collect all Host ID information.

tests/scale_up_status_updates/scale_up_status_updates_suite_test.go
```diff
@@ -1,71 +0,0 @@
...
-
-			step = "creating a datacenter resource with 2 racks/6 nodes"
-			k := kubectl.ApplyFiles(dcYaml)
-			ns.ExecAndLog(step, k)
-
-			ns.WaitForSuperUserUpserted(dcName, 600)
-
-			step = "checking that all nodes have been started"
-			nodeStatusesHostIds := ns.GetNodeStatusesHostIds(dcName)
-			Expect(len(nodeStatusesHostIds), 6)
```
johntrimble, line 57: This test isn't really testing anything to do with scale up (as the test name suggests), it is only checking that we record all the host ids. I don't think this warrants its own dedicated test due to how simple it is, much like how most of the Condition statuses don't warrant their own test. I've moved this check into tests/oss_test_all_the_things/oss_test_all_the_things_suite_test.go.

tests/oss_test_all_the_things/oss_test_all_the_things_suite_test.go
```diff
@@ -55,6 +55,12 @@ var _ = Describe(testName, func() {
 			k := kubectl.ApplyFiles(dcYaml)
 			ns.ExecAndLog(step, k)
 
+			ns.WaitForSuperUserUpserted(dcName, 600)
+
+			step = "check recorded host IDs"
+			nodeStatusesHostIds := ns.GetNodeStatusesHostIds(dcName)
+			Expect(len(nodeStatusesHostIds), 3)
```
jimdickinson, line 62: can you put this into one (or more) of the scale up tests too?

johntrimble:
If you consider all integration test run attempts together, this PR has actually passed all the tests. Some of our tests are flaky making it difficult to get a run where everything passes.

PR #103, opened 2020-06-02
KO-384 Refactor out webhook and improve validation

eldondevat:
This makes the webhook slightly more precise, in that we validate the namespace of the service that is used for the webhook is actually our namespace. I also added documentation on the webhook in the developer documentation folder. New additional validation happens here : https://github.com/datastax/cass-operator/blob/81a86a8f0571c32261ef4181f3039d2758ebcc6c/operator/cmd/manager/webhook.go#L34 and here: https://github.com/datastax/cass-operator/blob/81a86a8f0571c32261ef4181f3039d2758ebcc6c/operator/cmd/manager/webhook.go#L117

### DIFF
--- a/docs/developer/validating_webhook.md
+++ b/docs/developer/validating_webhook.md
@@ -1,0 +1,36 @@
+## The validating webhook.
+
+The operator offers, and installs when possible, a validating webhook for
+related CRDs. The webhook is intended to provide checks of the validity of an
+update or create request, where there might be CRD-specific guardrails that are
+not readily checked by implicit CRD validity. Such checks include preventing
+renaming certain elements of the deployment, such as the the cassandra cluster
+or the racks, which are core to the identity of a cassandra cluster.
+
+Validating webhooks have specific requirements in kubernetes:
+* They must be served over TLS
+* The TLS service name where they are reached must match the subject of the certificate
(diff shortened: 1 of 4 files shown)


johntrimble:
@eldondevat Is this PR just moving code to a different file? I'm totally on board with that I just want to make sure I didn't miss something.

eldondevat:
Sorry @johntrimble I don't think this PR was really very precise before, and the single commit nature of it didn't make that easier, I've added some more detail. Thanks!

docs/developer/validating_webhook.md
```diff
@@ -0,0 +1,36 @@
...
+explicitly configured in the kubernetes validatingwebhookconfiguration object.
+
+The operator takes a progressive-enhancement approach to enabling this webhook,
+which is described as follows:
+
+The operator will look for, and if present, use, the certificates in the
+default location that the controller-manager expects the certificates.  If the
+files there don't exist, or the certificate does not appear to be valid, then
+the operator will generate a self-signed CA, and attempt to update the various
+kuberentes references to that certificate, specifically:
```
jimdickinson, line 23:
```suggestion
kubernetes references to that certificate, specifically:
```

PR #104, opened 2020-06-02
Mage target to install helm similar to k3d target

eldondevat:
The version here was pulled from the release script.

### DIFF
--- a/mage/k8s/lib.go
+++ b/mage/k8s/lib.go
@@ -266,6 +266,21 @@ func Env() {
 func InstallTool() {
 	loadClusterSettings()
 	clusterActions.InstallTool()
+	installHelm()
+}
+
+func installHelm() {
+	cwd, err := os.Getwd()
+	mageutil.PanicOnError(err)
+	os.Chdir("/tmp")
+	shutil.RunVPanic("curl",
+		"https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3",
+		"-o", "helm_install.sh",
+	)
+	os.Setenv("DESIRED_VERSION", "v3.1.2")
+	os.Chmod("./helm_install.sh", 0755)
+	shutil.RunVPanic("./helm_install.sh")
+	os.Chdir(cwd)
 }
 
 // Configure kubectl to point to the specified cluster


mage/k8s/lib.go
```diff
@@ -266,6 +266,21 @@ func Env() {
 func InstallTool() {
 	loadClusterSettings()
 	clusterActions.InstallTool()
+	installHelm()
+}
+
+func installHelm() {
```
sandoichi, line 272: I think that this function should be moved into `mage/helm/lib.go` that already exists

mage/k8s/lib.go
```diff
@@ -266,6 +266,21 @@ func Env() {
 func InstallTool() {
 	loadClusterSettings()
 	clusterActions.InstallTool()
+	installHelm()
```
sandoichi, line 269: I'm not sure if it makes sense to always download and install helm every time we install the cli tool for kind/k3d/gcloud/etc?

If you move the `installHelm` function below into the `mage/helm/lib.go package ` (maybe also rename the func to `InstallTool`), and then make it a public function, we can add an entry to the root `magefile` to expose it as a top level target so that you can run `mage helm:InstallTool`

I do understand the thought process of "we need a way to ensure all tools at play are installed and using a supported version" but maybe that becomes another mage target entirely? Like a `mage ensureAllTools` or something :thinking:

eldondevat: > I'm not sure if it makes sense to always download and install helm every time we install the cli tool for kind/k3d/gcloud/etc?

FWIW, I believe the bash file this downloads doesn't replace helm if the required version already exists. It could also be good to look for the binary in the executable path ahead of time. It sometimes feels to me like we have a lot of mage targets, but if you want more, I can add more.

mage/k8s/lib.go
```diff
@@ -266,6 +266,21 @@ func Env() {
...
+
+func installHelm() {
+	cwd, err := os.Getwd()
+	mageutil.PanicOnError(err)
+	os.Chdir("/tmp")
+	shutil.RunVPanic("curl",
+		"https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3",
+		"-o", "helm_install.sh",
+	)
+	os.Setenv("DESIRED_VERSION", "v3.1.2")
```
jimdickinson, line 280:
```suggestion
	os.Setenv("DESIRED_VERSION", "v3.2.4")
```

jimdickinson:
poke @eldondevat @sandoichi do we want to merge this, or throw it away?

PR #105, opened 2020-06-02
Cleaning with linter

jimdickinson:
Running through the output of golangci-lint and gofumpt.


mage/kubectl/lib.go
```diff
@@ -62,9 +62,7 @@ func (k KCmd) ToCliArgs() []string {
 		args = append(args, fmt.Sprintf("--%s=%s", k, v))
 	}
 	args = append(args, k.Command)
-	for _, r := range k.Args {
-		args = append(args, r)
-	}
+	args = append(args, k.Args...)
```
respringer, line 65: I didn't know golang had array destructuring with ...

operator/pkg/httphelper/client.go
```diff
@@ -191,7 +191,7 @@ func (client *NodeMgmtClient) CallKeyspaceCleanupEndpoint(pod *corev1.Pod, jobs
 	}
 
 	request := nodeMgmtRequest{
-		endpoint: fmt.Sprintf("/api/v0/ops/keyspace/cleanup"),
+		endpoint: "/api/v0/ops/keyspace/cleanup",
```
johntrimble, line 194: 🤣

PR #109, opened 2020-06-08
KO-353 user password update

johntrimble:
Add dynamic watch of secrets so that relevant cassandra user can be updated upon secret change.


operator/pkg/utils/utilities.go
```diff
@@ -34,3 +34,37 @@ func SearchMap(mapToSearch map[string]interface{}, key string) map[string]interf
 
 	return make(map[string]interface{})
 }
+
+func FindValueIndexFromStringArray(a []string, v string) int {
```
jimdickinson, line 38: `IndexOf`?

operator/pkg/dynamicwatch/watch.go
```diff
@@ -0,0 +1,337 @@
...
+	FindWatchers(meta metav1.Object, object runtime.Object) []types.NamespacedName
+}
+
+type DynamicWatchesAnnotationImpl struct {
+	Client          runtimeClient.Client
+	Ctx             context.Context
+	WatchedType     metav1.TypeMeta
+	WatchedListType metav1.TypeMeta
+	Logger          logr.Logger
+}
```
johntrimble, line 42: Note that this can be used to watch any resource type, not just secrets.

operator/pkg/dynamicwatch/watch.go
```diff
@@ -0,0 +1,337 @@
...
+	value, ok := labels[WatchedLabel]
+	return ok && "true" == value
+}
+
+//
+// DAO functions
+//
+
+func (impl *DynamicWatchesAnnotationImpl) listAllWatched(namespace string) ([]unstructured.Unstructured, error) {
+	watchedList := &unstructured.UnstructuredList{}
```
johntrimble, line 206: I know, the big scary frak-your-static-typing `UnstructuredList`. It doesn't matter. We aren't digging through any maps. We do everything through the `metav1.Object` interface, because we really only care about the resource's metadata.

operator/pkg/reconciliation/handler.go
```diff
@@ -192,9 +194,12 @@ func (rc *ReconciliationContext) isValid(dc *api.CassandraDatacenter) error {
 
 // NewReconciler returns a new reconcile.Reconciler
 func NewReconciler(mgr manager.Manager) reconcile.Reconciler {
+	client := mgr.GetClient()
+	dynamicWatches := dynamicwatch.NewDynamicSecretWatches(client)
```
johntrimble, line 198: This instance needs to be available both during reconciliation _and_ when we set up our watches. Consequently, we put it on the `ReconcileCassandraDatacenter` and will pass along the reference to the `ReconciliationContext` when we create it.

operator/pkg/reconciliation/reconcile_datacenter.go
```diff
@@ -25,6 +26,14 @@ func (rc *ReconciliationContext) ProcessDeletion() result.ReconcileResult {
 		return result.Error(err)
 	}
 
+	// Clean up annotation litter on the user Secrets
+	err := rc.SecretWatches.RemoveWatcher(types.NamespacedName{
+		Name: rc.Datacenter.GetName(), Namespace: rc.Datacenter.GetNamespace(),})
+
+	if err != nil {
+		rc.ReqLogger.Error(err, "Failed to remove dynamic secret watches for CassandraDatacenter")
```
johntrimble, line 34: Looking for feedback. Should we return an error here or just log it? It seems wrong to hold up a delete by returning an error, but maybe that is the right thing to do.

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -703,6 +702,32 @@ func (rc *ReconciliationContext) createUser(user api.CassandraUser) error {
...
+func (rc *ReconciliationContext) UpdateSecretWatches() error {
+	dc := rc.Datacenter
+	users := rc.GetUsers()
+	names := []types.NamespacedName{}
+	for _, user := range users {
+		name := types.NamespacedName{Name: user.SecretName, Namespace: dc.Namespace}
+		names = append(names, name)
+	}
+	dcNamespacedName := types.NamespacedName{Name: dc.Name, Namespace: dc.Namespace,}
+	err := rc.SecretWatches.UpdateWatch(dcNamespacedName, names)
```
johntrimble, line 726: As an aside, if we wanted to, we could implement the dynamic watches without using annotations. Thats because this call here in reconciliation provides us with all the state we need, so it need not be persisted someplace else.

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -764,40 +786,6 @@ func findHostIdForIpFromEndpointsData(endpointsData []httphelper.EndpointState,
...
-	for _, v := range values {
-		idx := findValueIndexFromStringArray(a, v)
-		if idx < 0 {
-			// array does not contain this value, so add it
-			a = append(a, v)
-		}
-	}
-	return a
-}
-
```
johntrimble, line 800: These functions were moved to utils.

operator/pkg/utils/utilities.go
johntrimble: Yeah, we might be able to replace this. This function is not new, it was moved here from reconcile_racks.go.

operator/pkg/reconciliation/reconcile_datacenter.go
respringer: Yeah, let's just log it.  Otherwise, how would someone remove a dc that has some sort of error that prevents removing the watch, right?

operator/pkg/reconciliation/reconcile_racks.go
respringer: Hmmm.  Would that be a win, or does the annotation buy us anything?

johntrimble: Jim brought up that we really don't do anything like that at present in the operator.

operator/pkg/reconciliation/handler.go
```diff
@@ -68,9 +69,10 @@ var log = logf.Log.WithName("reconciliation_handler")
 type ReconcileCassandraDatacenter struct {
 	// This client, initialized using mgr.client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client   client.Client
-	scheme   *runtime.Scheme
-	recorder record.EventRecorder
+	client        client.Client
+	scheme        *runtime.Scheme
+	recorder      record.EventRecorder
+	SecretWatches dynamicwatch.DynamicWatches
```
jimdickinson, line 75: why is this field exported? can it not be?

operator/pkg/reconciliation/reconcile_datacenter.go
jimdickinson: I'm not so sure, but I'm fine with trying this as you've coded it up

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -676,7 +676,7 @@ func shouldUpsertUsers(dc api.CassandraDatacenter) bool {
 	return time.Now().After(lastCreated.Add(time.Minute * 4))
 }
 
-func (rc *ReconciliationContext) createUser(user api.CassandraUser) error {
+func (rc *ReconciliationContext) upsertUser(user api.CassandraUser) error {
```
jimdickinson, line 679: 👍

tests/superuser-secret-provided/superuser_secret_provided_test.go
```diff
@@ -113,6 +118,35 @@ var _ = Describe(testName, func() {
...
+				"--password", superuserNewPass,
+				"-e", "select * from system_schema.keyspaces;").
+				WithFlag("container", "cassandra")
+			ns.ExecAndLog(step, k)
+
+			step = "delete datacenter"
+			k = kubectl.Delete("CassandraDatacenter", dcName)
+			ns.ExecAndLog(step, k)
+
+			// Ensure secret annotations and labels cleaned up on DC delete
```
jimdickinson, line 142: 👍

operator/pkg/reconciliation/reconcile_racks.go
jimdickinson: 💯

operator/pkg/utils/utilities.go
johntrimble: There appears to be no generic IndexOf function, unless I've missed something.

operator/pkg/reconciliation/handler.go
johntrimble: Yes it must be exported. It's used in package `cassandradatacenter` where we set up the watches in the controller.

operator/pkg/utils/utilities.go
jimdickinson: I was suggesting we rename it

johntrimble: Would you be content with `IndexOfString`? I don't want a generic name for a non-generic function in a generic package.

operator/pkg/reconciliation/reconcile_racks.go
jimdickinson: yeah did we see examples of any other operators building up state in memory?

operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
```diff
@@ -110,6 +112,33 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Setup watches for Secrets. These secrets are often not owned by or created by
+	// the operator, so we must create a mapping back to the appropriate datacenters.
+
+	rd, ok := r.(*reconciliation.ReconcileCassandraDatacenter)
+	if !ok {
+		// This should never happen. - John 06/10/2020
```
jimdickinson, line 120: 🤣

operator/pkg/utils/utilities.go
jimdickinson: 👍

operator/pkg/reconciliation/reconcile_racks.go
johntrimble: Yes, elastic operator does this all in memory. It's a bit convoluted (because they have other stuff going on) but it is all in memory.

operator/pkg/reconciliation/handler.go
johntrimble: TODO: Add comment about why this is here.

operator/pkg/reconciliation/reconcile_racks.go
johntrimble: Specifically: https://github.com/elastic/cloud-on-k8s/blob/master/pkg/controller/common/watches/handler.go

It's a bit harder to follow, but they do the same thing, except providing individual event handlers, whereas I'm returning the NamespacedName to use when queueing a reconciliation request (which I think is sufficient for our needs). They update these registrations during reconciliation just like we do.

PR #112, opened 2020-06-09
use the PodIP for mgmt api calls

jsanda:
The `BuildPodHostFromPod` function returns the hostname that is used to make a request to the mgmt api. That hostname is managed by the StatefulSets headless service, which is the datacenter service. 

This does not alway work though because hostnames are not available until pods are ready, and there are times that the operator needs to make mgmt api requests when pods are not ready. I ran into this while working on support for a Reaper sidecar (ticket yet to be created). The Reaper container won't be ready until Cassandra is up and running. In this scenario the only reliable option is to use the pod IP for mgmt api requests.

### DIFF
--- a/operator/pkg/httphelper/client.go
+++ b/operator/pkg/httphelper/client.go
@@ -66,11 +66,22 @@ type CassMetadataEndpoints struct {
 	Entity []EndpointState `json:"entity"`
 }
 
-func BuildPodHostFromPod(pod *corev1.Pod) string {
+type NoPodIPError error
+
+func newNoPodIPError(pod *corev1.Pod) NoPodIPError {
+	return fmt.Errorf("pod %s has no IP", pod.Name)
+}
+
+func BuildPodHostFromPod(pod *corev1.Pod) (string, error) {
 	// This function previously returned the dns hostname which includes the StatefulSet's headless service,
 	// which is the datacenter service. There are times though that we want to make a mgmt api call to the pod
 	// before the dns hostnames are available. It is therefore more reliable to simply use the PodIP.
-	return pod.Status.PodIP
+
+	if len(pod.Status.PodIP) == 0 {
+		return "", newNoPodIPError(pod)
+	}
+
+	return pod.Status.PodIP, nil
 }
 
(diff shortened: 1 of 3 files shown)


operator/pkg/httphelper/client.go
```diff
@@ -68,17 +67,10 @@ type CassMetadataEndpoints struct {
...
 
 func BuildPodHostFromPod(pod *corev1.Pod) string {
-	return GetPodHost(
-		pod.Name,
-		pod.Labels[api.ClusterLabel],
-		pod.Labels[api.DatacenterLabel],
-		pod.Namespace)
-}
-
-func GetPodHost(podName, clusterName, dcName, namespace string) string {
```
jimdickinson, line 78: let's keep this around until we're sure we won't need it

operator/pkg/httphelper/client.go
```diff
@@ -68,11 +67,10 @@ type CassMetadataEndpoints struct {
...
 func BuildPodHostFromPod(pod *corev1.Pod) string {
-	return GetPodHost(
-		pod.Name,
-		pod.Labels[api.ClusterLabel],
-		pod.Labels[api.DatacenterLabel],
-		pod.Namespace)
+	// This function previously returned the dns hostname which includes the StatefulSet's headless service,
+	// which is the datacenter service. There are times though that we want to make a mgmt api call to the pod
+	// before the dns hostnames are available. It is therefore more reliable to simply use the PodIP.
+	return pod.Status.PodIP
```
jimdickinson, line 73: if this returns the empty string, what will happen?

jsanda: I am not sure under what conditions this would return an empty string. Do you want me to add a check and return an error if it is the empty string?

jimdickinson: I tried checking this on Pending / Terminating pods, and it looks like there are some windows where there won't be a PodP.

Adding an error return value changes the signature and makes this a bigger fix, but that might be the best way to handle this.

jsanda: I updated the function to return an error when there is no IP.

Issue #113, opened 2020-06-09
Add support for Reaper


jsanda:
I propose adding support for running repairs with [Reaper](http://cassandra-reaper.io) by introducing a Reaper sidecar. Reaper already has a [sidecar mode](http://cassandra-reaper.io/docs/usage/sidecar_mode/) which lends itself really well to this. 

High level goals for initial integration include:

* Run Reaper as a sidecar container
* Make deploying Reaper optional
* Provide ability to add/remove Reaper after `CassandraDatacenter` has already been created
* Make Reaper UI accessible through a service
    * Reaper users rely heavily on using the UI for managing repairs
* Expose minimal Reaper configs through `CassandraDatacenter` CRD
* Enable Reaper authentication by default and provide (Reaper) admin credentials in none specified

An example manifest might look like this:

```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: cluster1
  serverType: cassandra
  serverVersion: "3.11.6"
  managementApiAuth:
    insecure: {}
  size: 3
  racks:
    - name: rack1
    - name: rack2
    - name: rack3
  reaper:
    enabled: true
    adminUserSecretName: reaperAdmin
```

When Reaper is enabled this would result in the following:

* Deploy Reaper sidecar container in Cassandra pod
* Initialize `reaper_db` keyspace
* Add Cassandra cluster to Reaper (done automatically when using Reaper sidecar mode)
* Expose Reaper UI via service

What is not in scope for this ticket?

* Exposing any and all Reaper configuration
    * We want to expose as little configuration as possible and provide sensible defaults
* Running Reaper in a mode other than sidecar
    * This would be better suited for stand alone Reaper operator I am working on, [reaper-operator](https://github.com/thelastpickle/reaper-operator)

jimdickinson:
Reaper is managed as part of k8ssandra.io now

PR #116, opened 2020-06-10
Reaper sidecar

jsanda:
This is for #113.

Here is what has been done so far:

* Reaper sidecar container is deployed
* Reaper schema initialized via k8s job
* Reaper UI exposed via k8s service

Here are tasks still outstanding:

* Make Reaper deployment conditional via CRD
* Add unit/integration tests
* Turn on Reaper authentication by default and create default admin user if necessary


jsanda:
@jimdickinson This is ready for review.

operator/pkg/httphelper/client.go
```diff
@@ -67,12 +66,22 @@ type CassMetadataEndpoints struct {
...
-func BuildPodHostFromPod(pod *corev1.Pod) string {
-	return GetPodHost(
-		pod.Name,
-		pod.Labels[api.ClusterLabel],
-		pod.Labels[api.DatacenterLabel],
-		pod.Namespace)
+type NoPodIPError error
+
+func newNoPodIPError(pod *corev1.Pod) NoPodIPError {
+	return fmt.Errorf("pod %s has no IP", pod.Name)
```
jimdickinson, line 72: if you merge master in, some of this would go away I think?

operator/pkg/reconciliation/constructor.go
```diff
@@ -379,6 +379,7 @@ func buildContainers(dc *api.CassandraDatacenter, serverVolumeMounts []corev1.Vo
 		{Name: "MGMT_API_EXPLICIT_START", Value: "true"},
 		// TODO remove this post 1.0
 		{Name: "DSE_MGMT_EXPLICIT_START", Value: "true"},
+		{Name: "IGNORE_DEFAULTS", Value: "true"},
```
jimdickinson, line 382: this needs to go in #129 right?

operator/pkg/reconciliation/constructor.go
```diff
@@ -410,7 +411,13 @@ func buildContainers(dc *api.CassandraDatacenter, serverVolumeMounts []corev1.Vo
 	}
 	loggerContainer.VolumeMounts = []corev1.VolumeMount{cassServerLogsMount}
 
-	return []corev1.Container{cassContainer, loggerContainer}, nil
+	containers := []corev1.Container{cassContainer, loggerContainer}
+	if dc.Spec.Reaper.Enabled {
```
jimdickinson, line 415: should we add that `dc.Spec.ServerType == "cassandra"` also?

operator/pkg/reconciliation/reconcile_reaper.go
```diff
@@ -0,0 +1,235 @@
...
+	} else if err != nil {
+		return result.Error(err)
+	} else if jobFinished(schemaJob) {
+		return result.Continue()
+	} else {
+		return result.RequeueSoon(2)
+	}
+}
+
+func buildInitReaperSchemaJob(dc *api.CassandraDatacenter) *v1batch.Job {
```
jimdickinson, line 113: move this over to the constructor.go file?

operator/pkg/reconciliation/reconcile_reaper.go
```diff
@@ -0,0 +1,235 @@
...
+const (
+	ReaperUIPort             = 7080
+	ReaperAdminPort          = 7081
+	ReaperDefaultImage     = "thelastpickle/cassandra-reaper:2.0.5"
+	ReaperDefaultPullPolicy  = corev1.PullIfNotPresent
+	ReaperContainerName      = "reaper"
+	ReaperHealthCheckPath    = "/healthcheck"
+	ReaperKeyspace           = "reaper_db"
+	ReaperSchemaInitJob      = "ReaperSchemaInitJob"
+	ReaperSchemaInitJobImage = "jsanda/reaper-init-keyspace:latest"
```
johntrimble, line 29: This probably needs to be updated? Where does this init-container live at present?

operator/pkg/reconciliation/reconcile_reaper.go
```diff
@@ -0,0 +1,235 @@
...
+		},
+		Spec: v1batch.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					RestartPolicy: corev1.RestartPolicyOnFailure,
+					Containers: []corev1.Container{
+						{
+							Name:getReaperSchemaInitJobName(dc),
+							Image: ReaperSchemaInitJobImage,
+							ImagePullPolicy: corev1.PullAlways,
```
johntrimble, line 132: I'm guessing this is leftover from debugging?

operator/pkg/reconciliation/reconcile_reaper.go
```diff
@@ -0,0 +1,235 @@
...
+							Env: []corev1.EnvVar{
+								{
+									Name: "KEYSPACE",
+									Value: ReaperKeyspace,
+								},
+								{
+									Name: "CONTACT_POINTS",
+									Value: dc.GetSeedServiceName(),
```
johntrimble, line 140: So this is a DNS name that can be resolved to all the seeds in the datacenter. Does Reaper need every seed or just a single seed? I ask, because by default I'd assume it would resolve the DNS to just a single IP.

johntrimble: Also, how does it handle seeds changing over time?

operator/pkg/reconciliation/constructor.go
jsanda: yeah, I just added it in #129.

johntrimble: Weren't we going to update the management api to default the value of this to that of `USE_MGMT_API` to retain compatibility?

operator/pkg/reconciliation/reconcile_reaper.go
jsanda: It currently lives at https://github.com/jsanda/create_keyspace. It is deployed as a job, not as an init container. The schema has to be initialized in a job in order to prevent concurrent schema modifications.

Once the management API supports creating keyspaces and tables, this won't be necessary.

operator/pkg/reconciliation/reconcile_reaper.go
jsanda: >I ask, because by default I'd assume it would resolve the DNS to just a single IP

For a headless service I believe it gets mapped to all endpoints.

>how does it handle seeds changing over time?

I added a patch to Reaper a while back to handle C* node address changes. The behavior for this is controlled by the `REAPER_ENABLE_DYNAMIC_SEED_LIST` env var. Also note that Reaper does not care about seed nodes per se. It just needs an initial contact point, and from there it will discover other nodes via JMX.

jsanda:
I rebased off of master and made changes based on reviews.

operator/pkg/reconciliation/reconcile_reaper.go
jsanda: I moved the `buildInitReaperSchemaJob` function. Did you want the `buildReaperContainer` function moved as well?

operator/pkg/reconciliation/constructor.go
jsanda: done

operator/pkg/reconciliation/reconcile_reaper.go
jsanda: fixed

operator/pkg/reconciliation/reconcile_reaper.go
jsanda: @johntrimble I may have misunderstood your initial point/question about seeds. Reaper only needs a single contact point; however, it is better to provide more than one. Reaper will go through the set of contact points until it establishes a successful connection.

operator/pkg/reconciliation/reconcile_reaper.go
jsanda: >Have we run integration tests against this branch?

I have not run all integration tests locally. I just committed/pushed an integration that I just realized hadn't been committed.

johntrimble: How do we build new versions of this image and deploy them? Should it be just another image built by the cass operator build process (we already have more than one image we build).

jsanda: That is a good question. I personally lean towards keeping it separate because I am using it for a stand-alone operator for Reaper as well. With that said, hopefully this can altogether replaced with calls to the mgmt API in the near future so I am open to whatever makes the most sense in the interim.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -50,7 +50,7 @@ type ProgressState string
 // In the event that no image is found, an error is returned
 func getImageForServerVersion(server, version string) (string, error) {
 	const (
-		cassandra_3_11_6 = "datastax/cassandra-mgmtapi-3_11_6:v0.1.2"
+		cassandra_3_11_6 = "datastax/cassandra-mgmtapi-3_11_6:v0.1.4"
```
jimdickinson, line 53: if you merge master in, I think this will go away

operator/pkg/reconciliation/constructor.go
jimdickinson: with 0.1.5, this is no longer needed

operator/pkg/reconciliation/reconcile_reaper.go
jimdickinson: I think leaving this as-is is fine for merging this PR

operator/pkg/reconciliation/constructor.go
jsanda: I removed the env var

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
jsanda: I rebased off master and updated to v0.1.5.

Issue #120, opened 2020-06-13
StatefulSet can be orphaned after podTemplateSpec update


jsanda:
**Summary**
If you update `.spec.podTemplateSpec` and then subsequently delete the `CassandraDatacenter`, the Statefulset does not get deleted. It looks like this is caused by the controller reference not getting set on the updated StatefulSet after `podTemplateSpec` is updated.

This is happening in master.

**Steps to Reproduce**
Deploy `CassandraDatacenter`:

```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: test-cluster
  serverType: cassandra
  serverVersion: 3.11.6
  managementApiAuth:
    insecure: {}
  size: 3
  storageConfig:
    cassandraDataVolumeClaimSpec:
      storageClassName: server
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
``` 

After the `dc1` is ready, update the `podTemplateSpec`:

```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: test-cluster
  serverType: cassandra
  serverVersion: 3.11.6
  managementApiAuth:
    insecure: {}
  size: 3
  storageConfig:
    cassandraDataVolumeClaimSpec:
      storageClassName: server
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
  podTemplateSpec:
    spec:
      containers: []
      terminationGracePeriodSeconds: 31
```

Delete `dc1` and there is no cascading delete for the StatefulSet,`test-cluster-dc1-default-sts`.

jimdickinson:
Great find!

jimdickinson:
Fixed by #121

PR #121, opened 2020-06-13
set controller reference after PodTemplateSpec update

jsanda:
#120 

This PR probably deserves a unit test and/or integration test. 

@jimdickinson I am happy to update the PR with tests, but I want to first make sure that the code change is ok.


johntrimble:
Looks great @jsanda !

jimdickinson:
Fixes #120

PR #122, opened 2020-06-15
Login example for README

### LINKED ISSUE #61: Add cqlsh login example to the README
We have something good in https://github.com/datastax/cass-operator/blob/master/docs/user/README.md#superuser-credentials

but no examples in the README.

jimdickinson:
Fixes #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,29 @@ UN  10.233.92.96    186.48 KiB  1            61.6%             b119eae5-2ff4-4b0
 UN  10.233.90.54    205.1 KiB   1            73.1%             0a96e814-dcf6-48b9-a2ca-663686c8a495  r1
 ```
 
+The operator creates a secure Cassandra cluster by default, with a new superuser (not the traditional `cassandra` user) and a random password. You can get those out of a Kubernetes secret and use them to log into your Cassandra cluster for the first time. For example:
+
+```console
+$ # get CASS_USER and CASS_PASS variables into the current shell
+$ CASS_USER=$(kubectl -n cass-operator get secret cluster1-superuser -o json | jq -r '.data.username' | base64 --decode)
+$ CASS_PASS=$(kubectl -n cass-operator get secret cluster1-superuser -o json | jq -r '.data.password' | base64 --decode)
+$ kubectl -n cass-operator exec -ti cluster1-dc1-default-sts-0 -c cassandra -- bash -c "cqlsh -u $CASS_USER -p $CASS_PASS"
+
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -130,6 +130,28 @@ UN  10.233.92.96    186.48 KiB  1            61.6%             b119eae5-2ff4-4b0
 UN  10.233.90.54    205.1 KiB   1            73.1%             0a96e814-dcf6-48b9-a2ca-663686c8a495  r1
 ```
 
+The operator creates a secure Cassandra cluster by default, with a new superuser (not the traditional `cassandra` user) and a random password. You can get those out of a Kubernetes secret and use them to log into your Cassandra cluster for the first time. For example:
+
+```console
+$ # get CASS_USER and CASS_PASS variables into the current shell
+$ CASS_USER=$(kubectl -n cass-operator get secret cluster1-superuser -o json | jq -r '.data.username' | base64 --decode)
+$ CASS_PASS=$(kubectl -n cass-operator get secret cluster1-superuser -o json | jq -r '.data.password' | base64 --decode)
+$ kubectl -n cass-operator exec -ti cluster1-dc1-default-sts-0 -c cassandra -- bash -c "cqlsh -u $CASS_USER -p $CASS_PASS"
```
johntrimble, line 139: You should be able to just do:

```
kubectl -n cass-operator exec -ti cluster1-dc1-default-sts-0 -c cassandra -- cqlsh -u $CASS_USER -p $CASS_PASS
```
There's no need to invoke bash here.

jimdickinson: I was thinking that too - do you think it'll be safe for all characters that can go in passwords? I know it should be good for ones we generate, I think

johntrimble: Oh, you should put quotes around `"$CASS_USER"` and `"$CASS_PASS"`, but I don't see how using bash helps you avoid special characters dorking things up.

johntrimble: Actually, if the password contains a `$` what you have won't work because the shell on the client will replace `$CASS_PASS` with password, and then bash in the container will interpret the `$` in the password as a variable and replace it.

johntrimble:
```$ X='$bob'
$ echo bob is $X
bob is $bob
$ docker run -it --rm ubuntu:20.04 /bin/bash -c "echo X is $X"
X is```

Comment (jimdickinson) [reply README.md:139]: I think the way to fix that is with single quotes, but this definitely starts to get hard to read quickly

Comment (jimdickinson) [reply README.md:139]:
```
$ docker run -it --rm ubuntu:20.04 /bin/bash -c "echo X is '$X'"
X is $bob
```

PR #123, opened 2020-06-15
FAQ

jimdickinson:


FAQ.md
```diff
@@ -0,0 +1,8 @@
+# Frequently Asked Questions
+
+## My pods are stuck in a Pending state. Why won't they come up?
+
+This one has a lot of potential answers...
+
+- `kubectl describe pod` and check the message for why the current nodes aren't being used.
+- Did you pin any `racks` to a `zone` in a different region, or one without any k8s workers?
```
johntrimble, line 8: I'm about to say something about punctuation. I'm not a punctuation expert. I like to season my prose with punctuations until it looks right. However, the comma before "or" is, I think, unnecessary. This isn't a conjunction of two sentences, and it isn't a list of more than two things. There may be some other rule that necessitates a comma though, so I'm going to approve this PR anyway.

bradfordcp:
Should this be in a draft state? I don't see why this is open pending a merge.

PR #129, opened 2020-06-19
bump version of mgmt sidecar to 0.1.5

jsanda:
v0.1.4 includes the fix for the [unix socket and TCP socket driver conflict issue](https://github.com/datastax/management-api-for-apache-cassandra/issues/18).

This fix is needed for #116 among other things.

### DIFF
--- a/buildsettings.yaml
+++ b/buildsettings.yaml
@@ -19,4 +19,4 @@ dev:
   images:
     - datastax/dse-server:6.8.0
     - datastax/cass-config-builder:1.0.0
-    - datastax/cassandra-mgmtapi-3_11_6:v0.1.2
+    - datastax/cassandra-mgmtapi-3_11_6:v0.1.5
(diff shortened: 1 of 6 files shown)


jimdickinson:
I think that environment variable change was needed here too? or would we rather release 0.1.5 using env vars we already send down to indicate "this has been setup by Cass Operator and the config-builder init container"?

jimdickinson:
Sorry for the thrashing - I just made this happen - https://github.com/datastax/management-api-for-apache-cassandra/releases/tag/v0.1.5 - so no need for env var changes

jimdickinson:
also there's this straggler at the bottom of buildsettings.yaml, which is just an optimization for less hitting the network for integration tests...

https://github.com/datastax/cass-operator/blob/151d4a0e18ba993f7308cec07e29666bcacb2739/buildsettings.yaml#L22

Issue #131, opened 2020-06-23
Support "Seed Hinting"


bradfordcp:
As we look to start supporting C* clusters that span K8s clusters we need to start adding some building blocks around the seeds list. Currently, cass-operator configures the seed list with the DNS name for the DC seed service. If a user specifies a seed entry (or entries) in their `spec.config.cassandra-yaml.seeds` YAML path we should not override it and simply append the value for this CRs DC.

bradfordcp:
It looks like the path I mentioned above `spec.config.cassandra-yaml.seeds` does NOT exist. The correct path would be `spec.config.cluster-info.seeds`. See https://github.com/datastax/cass-config-builder/blob/master/README.md#cluster-info-key for more information.

jimdickinson:
@bradfordcp do you think what we've got for `additionalSeeds` satisfies this?

bradfordcp:
Yep, closing

PR #136, opened 2020-06-23
KO-440 Provide an easy way for people to specify additional seeds in the crd

sandoichi:


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -165,6 +165,8 @@ type CassandraDatacenterSpec struct {
 
 	// Cassandra users to bootstrap
 	Users []CassandraUser `json:"users,omitempty"`
+
+	StaticSeedIPs []string `json:"staticSeedIPs,omitempty"`
```
jimdickinson, line 169: we should think hard about this name... @bradfordcp used the word "hint" 🤔 

https://github.com/datastax/cass-operator/issues/131

operator/pkg/apis/cassandra/v1beta1/seed_providers.go
```diff
@@ -0,0 +1,147 @@
...
+	switch parameters := parameters.(type) {
+	case []map[string]interface{}:
+		for _, m := range parameters {
+			if v, ok := m["seeds"]; ok {
+				switch v := v.(type) {
+				case []string:
+					seeds = v
+				}
+			} else {
+				otherParams = append(otherParams, m)
```
sandoichi, line 50: keep track of any other param maps other than seeds so we can preserve them when we do an update

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
bradfordcp: Do we want to specify seed entries at the top-level of the CRD or in something like `config.cluster-info.seeds` per https://github.com/datastax/cass-config-builder/blob/master/README.md#cluster-info-key?

bradfordcp: What happens if a user specifies both?

jimdickinson: I don't think we should be putting this stuff under `config` btw, and if they specify both we overwrite the config-blob one

operator/pkg/apis/cassandra/v1beta1/seed_providers.go
```diff
@@ -0,0 +1,151 @@
...
+	otherParams, seeds := findSeedsAndOtherParams(parameters)
+
+	var className string
+	if c, ok := sp["class_name"]; ok {
+		className = fmt.Sprintf("%v", c)
+	} else {
+		className = buildClassNameForSeedProvider(dc.Spec.ServerType)
+	}
+
+	return &SeedProvider{ClassName: className, Seeds: seeds, OtherParams: otherParams}
```
johntrimble, line 79: It doesn't look like we really care about the seed provider here. I think you just want to use the `seed` option we give to config builder as part of the `cluster-info`. See:

https://github.com/datastax/cass-operator/blob/master/operator/pkg/serverconfig/configgen.go#L14-L28

and

https://github.com/datastax/cass-operator/blob/248b30efe797d0656f2fc5c8e96dc3c431ab9a32/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go#L432-L435

This way we are letting config builder handle the details of figuring out the correct class name given the cassandra version.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -432,7 +434,9 @@ func (dc *CassandraDatacenter) GetConfigAsJSON() (string, error) {
 	// We use the cluster seed-service name here for the seed list as it will
 	// resolve to the seed nodes. This obviates the need to update the
 	// cassandra.yaml whenever the seed nodes change.
-	modelValues := serverconfig.GetModelValues([]string{dc.GetSeedServiceName()}, dc.Spec.ClusterName, dc.Name)
+	seeds := []string{dc.GetSeedServiceName()}
+	seeds = append(seeds, dc.Spec.StaticSeedIPs...)
+	modelValues := serverconfig.GetModelValues(seeds, dc.Spec.ClusterName, dc.Name)
```
johntrimble, line 439: 💯

jimdickinson:
probably want to change the PR title ;)

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
jimdickinson, line 169: I'm gonna suggest `AdditionalSeeds`

jimdickinson:
We should write out a test for this that uses two operators / namespaces / cassandra clusters in the same k8s cluster

PR #140, opened 2020-06-25
Fixes #51 - Added ingress documentation and sample client application to docs

### LINKED ISSUE #51: Document connecting to a cluster running within k8s from outside the k8s cluster
It's pretty straightforward to connect from within the cluster (using a headless service or DNS entry), but how do you connect from outside k8s?

bradfordcp:
Currently, the only reference implementation is Traefik, but I expoect to see multiple tickets coming up for the other ingresses that should be documented. For now I'm considering this done and ready for merging in to the mainline.


docs/proxy/README.md
```diff
@@ -0,0 +1,133 @@
...
+* If a pod is rescheduled the IP address of the pod can change
+* In some K8s distributions this is a privileged operation
+* Additional automation will be required to identify the appropriate IP and set it for listen_address and broadcast_address
+* Only one Cassandra pod may be started per worker, regardless of `allowMultiplePodsPerWorker` setting.
+
+### Host Port
+
+Host port is similar to host network, but instead of being applied at the pod level, it is applied to specified containers within the pod. For each port listed in the container's block a hostPort: external_port key value is included. external_port is the port number on the Kubernetes worker that should be forwarded to this container's port.
+
+At this time we do not allow for modifying the cassandra container via podTemplateSpec, thus configuring this value is not possible without patching each rack's stateful set.
```
johntrimble, line 46: Oi! We should probably fix that :-/

docs/proxy/traefik/mtls-sni/README.md
```diff
@@ -0,0 +1,114 @@
...
+    kubectl apply -f traefik/mtls-sni/sample-cluster-sample-dc.tlsoption.yaml
+    ```
+
+1. Edit and create the `IngressTCPRoute`. This provides the SNI mapping for routing TCP requests from the ingress to individual pods.
+
+    ```bash
+    kubectl apply -f traefik/mtls-sni/sample-cluster-sample-dc.ingressroutetcp.yaml
+    ```
+
+1. Create the `service` for the pod with `kubectl expose`. Note the service name will match the pod name.
```
johntrimble, line 106: I feel like there's an argument to be made here for having an option to get the operator to create a service per cassandra node.

docs/proxy/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+It is possible to configure a service within Kubernetes outside of those provided by cass-operator that is accessible from outside of the Kubernetes cluster. These services have a type: LoadBalancer key in the spec block. In most cloud environments this results in a native cloud load balancer being provisioned to point at the appropriate pods with an external IP. Once the load balancer is provisioned running kubectl get svc will display the external IP address that is pointed at the C* nodes.
+
+**Pros**
+
+* Available from outside of the cluster
+
+**Cons**
+
+* Requires WhiteListPolicy Load Balancing Policy (LBP) to restrict attempts by the drivers to connect directly with pods
```
johntrimble, line 76: I don't see an example of this. Might be nice in the future as this looks like a relatively straightforward way of just getting _something_ working.

docs/proxy/traefik/mtls-sni/README.md
bradfordcp: There is, some ingresses will allow you to specify the pod under a stateful set to route to instead of a service.

bradfordcp: See https://voyagermesh.com/docs/10.0.0/guides/ingress/http/statefulset-pod/#forward-traffic-to-specific-pods-of-a-statefulset

docs/proxy/README.md
bradfordcp: Heh, this needs to change a little as we removed the `WhiteListPolicy`. Now it's a simple implementation of `AddressTranslator` (see our [docs](https://docs.datastax.com/en/developer/java-driver/4.7/manual/core/address_resolution/#driver-side-address-translation) and my [example](https://github.com/datastax/cass-operator/blob/proxy-config/docs/proxy/sample-java-application/src/main/java/com/datastax/kubernetes/KubernetesIngressAddressTranslator.java).

Issue #127, opened 2020-06-17
Make Ergonomic and Easy automation for Internode Encryption


eldondevat:
Currently, we've discussed using additional, user-supplied initcontainers to provision certificates for internode encryption (as discussed in #108), but it would be nice to have an easy, ergonomic provisioning method for internode encryption. This could be based on a user-supplied CA, an operator-generated internal CA, or individually provisioned certificates. The purpose of this ticket is to get feedback on some of the strategies that have been suggested in ad-hoc discussions.

jimdickinson:
After looking into cert-manager a bit, I don't think it helps us much here. It would be great at getting the same cert on all of the Cassandra pods.

PR #156, opened 2020-07-06
Create JKS for internode encryption (#127)

eldondevat:
Automatically bootstrapping a certificate authority and enabling internode encryption.


operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1515,6 +1527,34 @@ func (rc *ReconciliationContext) findStartedNotReadyNodes() (bool, error) {
 	return false, nil
 }
 
+func (rc *ReconciliationContext) copyPodCredentials(pod *corev1.Pod, jksBlob []byte) error {
```
jimdickinson, line 1530: no pod stuff ended up getting used?

eldondevat: I think this belongs in the/an init container at this point. This gets us encryption without the fancy peer verification, it's true, but I would like to get this in so it can be iterated on.

operator/pkg/utils/crypto.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return "", "", err
+}
+
+func prepare_ca(ca *corev1.Secret) (ca_cert_bytes []byte, ca_certificate *x509.Certificate, ca_key *rsa.PrivateKey, err error) {
+	ca_certificate_pem, _ := pem.Decode(ca.Data["cert"])
+	ca_cert_bytes = ca_certificate_pem.Bytes
+	ca_key_block, _ := pem.Decode(ca.Data["key"])
+	if untyped_ca_key, ca_key_err := x509.ParsePKCS8PrivateKey(ca_key_block.Bytes); ca_key_err != nil {
+		err = ca_key_err
```
respringer, line 77: err is not being passed in by reference.  I don't think the value of err in the caller will be effected unless you pass "err *error" as the parameter.  In any case, why aren't we just returning error instead of modifying the parameter?

tests/internode_encryption/internode_encryption_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	ginkgo_util "github.com/datastax/cass-operator/mage/ginkgo"
+	"github.com/datastax/cass-operator/mage/kubectl"
+)
+
+var (
+	testName     = "Add racks"
```
respringer, line 18: We need to update the testName and the test namespace name.

tests/internode_encryption/internode_encryption_test.go
```diff
@@ -0,0 +1,118 @@
...
+		ns.Terminate()
+	})
+
+	RegisterFailHandler(Fail)
+	RunSpecs(t, testName)
+}
+
+var _ = Describe(testName, func() {
+	Context("when in a new cluster", func() {
+		Specify("racks can be added if the size is increased accordingly", func() {
```
respringer, line 42: Also this message needs to be updated.

operator/pkg/utils/crypto.go
eldondevat: Err is not being passed in at all. Err is a [named return](https://tour.golang.org/basics/7)

tests/internode_encryption/internode_encryption_test.go
eldondevat: Thanks, good catch.

tests/internode_encryption/internode_encryption_test.go
eldondevat: Yep, will update. Thanks!

eldondevat:
Updates pushed, the above test run just appears to be flaky integration tests?

operator/patchy
```diff
@@ -0,0 +1,81 @@
+diff --git a/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go b/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```
jimdickinson, line 1: did you need this in the PR?

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -663,11 +663,11 @@ func (dc *CassandraDatacenter) GetContainerPorts() ([]corev1.ContainerPort, erro
 			ContainerPort: 8609,
 		},
 		{
-			Name:          "intra-node",
+			Name:          "inter-node",
```
jimdickinson, line 666: "internode" seems more common than "inter-node"

operator/pkg/reconciliation/reconcile_racks.go
jimdickinson: 👍

jimdickinson:
Not a blocker, but is there a good path to optionally putting the CA in ahead of time so we can get two DCs orchestrated to work together?

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -663,11 +663,11 @@ func (dc *CassandraDatacenter) GetContainerPorts() ([]corev1.ContainerPort, erro
 			ContainerPort: 8609,
```
jimdickinson, line 663: I think we should put a `tls` toggle on the cassdc, maybe with a secret name input too

operator/patchy
eldondevat: No, sorry!

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
eldondevat: Ok, that's not a change I made, but happy to do so. Can this be a future PR?

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
eldondevat: Yeah, I agree. Will update.

operator/pkg/utils/crypto.go
respringer: oh, ok.  TIL

eldondevat:
> Not a blocker, but is there a good path to optionally putting the CA in ahead of time so we can get two DCs orchestrated to work together?

In this case, it is all stored as secrets, and should use the secrets present if they are available. So generating the CA ahead of time and populating the secret I think should work. Still, that strategy deserves an integration test and additional developer/user documentation, which I realize is sadly lacking from this PR. I think a follow-on PR makes sense to address a lot of these.

eldondevat:
We need to add docs/tests around the following scenarios:
* User doc about how to configure the thing
* How to use a single CA across multiple cassdc's
* What happens when this gets flipped from "off" to "on"

PR #159, opened 2020-07-13
Ko 461 internode node port service

respringer:
This PR adds Spec.Networking.NodePortConfig that contains CqlPort and BroadcastPort.  If these two settings are defined then the cql service will preserve source IPs and be configured as a NodePort service.

The specified port values will be passed to config-builder, along with the k8s node IP as "HOST_IP".  Additionally a boolean environment variable called "USE_HOST_IP_FOR_BROADCAST" will be set to "true" and given to config builder if NodePortConfig is defined.


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -219,6 +219,9 @@ type CassandraDatacenterSpec struct {
 	// Cassandra users to bootstrap
 	Users []CassandraUser `json:"users,omitempty"`
 
+	// Preserve client source-ips
+	PreserveClientSourceIps bool `json:"preserveClientSourceIps,omitempty"`
```
jimdickinson, line 223: maybe something like

```suggestion
	NodePort *int `json:"nodePort,omitempty"`
```

where they can set it to 0 to get an auto-assigned port, or any number to request a specific port?

jimdickinson: oh maybe we need two ports on the node port service? 🤦 one to replace 9042 and another to replace 7000 (or 7001)

jimdickinson:
I think we need something to push down the broadcast address and ports grabbed

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -221,11 +221,29 @@ type CassandraDatacenterSpec struct {
...
 
+type NetworkingConfig struct {
+	NodePort *NodePortConfig `json:"nodePort,omitempty"`
+}
+
+type NodePortConfig struct {
+	CqlPort          int `json:"cqlPort,omitempty"`
+	CqlSslPort       int `json:"cqlSslPort,omitempty"`
+	BroadcastPort    int `json:"broadcastPort,omitempty"`
+	BroadcastSslPort int `json:"broadcastSslPort,omitempty"`
```
jimdickinson, line 239: maybe remove the word `Port` from the field names here - a JSON path would look like `networking.NodePort.CqlPort` right now?

PR #119, opened 2020-06-12
updated operator to work with multiple namespaces

rltvty:
split the startup into pulling 2 namespaces... `operatorNs` for the namespace the operator is running in, and `watchNs` for the namespaces passed via the `WATCH_NAMESPACE` environment variable.

I've successfully tested this in my development EKS cluster.  I'm able to get the operator to deploy Cassandra datacenters in namespaces outside the operator.

Issue #115, opened 2020-06-10
validation fails if multiple watch namespaces


ninjaslothx:
If modifying the WATCH_NAMESPACES var ensureWebhookConfigVolume and ensureWebhookCertificate fails.  

The namespace var is imported from the WATCH_NAMESPACE env var here:
https://github.com/datastax/cass-operator/blob/master/operator/cmd/manager/main.go#L97

While the use of multiple namespaces has been accounted for here https://github.com/datastax/cass-operator/blob/master/operator/cmd/manager/main.go#L147

The two validations before this cause the container to fail. 
https://github.com/datastax/cass-operator/blob/master/operator/cmd/manager/main.go#L128

It might be better to add an additional env var e.g OPERATOR_NAMESPACE to read in and use here, setting it to the value of the operators namespace specifically. 
Then use the watch_namespace var purely for its intended function.

jimdickinson:
Thanks for the bug report! I think there's an established way to get the operator's runtime namespace here...

https://github.com/operator-framework/operator-sdk/blob/master/pkg/k8sutil/k8sutil.go#L64

So I think we can fix this without an extra env var. Thinking about this a little more, RBAC will probably stop the operator from creating resources in another namespace with our stock configuration, but it's just about changing the Role to a ClusterRole (and RoleBinding to a ClusterRoleBinding). How are you planning on using the operator w.r.t. namespaces?

rltvty:
I think I encountered this RBAC issue.  After clearing the `WATCH_NAMESPACE` var, I'm getting the following from the deployment logs: 

```
E0611 22:46:13.776329       1 reflector.go:153] pkg/mod/k8s.io/client-go@v0.17.4/tools/cache/reflector.go:105: Failed to list *v1beta1.CassandraDatacenter: cassandradatacenters.cassandra.datastax.com is forbidden: User "system:serviceaccount:cass-operator:cass-operator" cannot list resource "cassandradatacenters" in API group "cassandra.datastax.com" at the cluster scope
```

I'd like to deploy the operator in a single namespace, and then be able to create Cassandra datacenters in multiple namespaces.  There may be many namespaces where I do this, so I'd like to avoid having to deploy the operator to each namespace first.

rltvty:
I had to update the `clusterRole` to the following to get this to work:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  creationTimestamp: null
  name: cass-operator-cluster-role
rules:
[...]

PR #173, opened 2020-07-21
Use k8sutils directly instead of passing unnecessary parameters

eldondevat:
This should close (#115) (#119)


jsanda:
I rebased off of v1.3.0 and tested this. It works nicely! I can write up a short blog post summarizing the changes I made to the stock configuration to get the operator to watch two namespaces.

PR #175, opened 2020-07-23
Headless ClusterIP service for additional seeds

respringer:
This adds a headless ClusterIP service for the additional seeds.  The additional seeds must be specified as IP addresses, and not as hostnames.  The additional seeds ip addresses are added to an Endpoints resource.

### DIFF
--- a/operator/pkg/reconciliation/reconcile_endpoints.go
+++ b/operator/pkg/reconciliation/reconcile_endpoints.go
@@ -49,8 +49,6 @@ func (rc *ReconciliationContext) CheckAdditionalSeedEndpoints() result.Reconcile
 
 	logger.Info("reconcile_endpoints::CheckAdditionalSeedEndpoints")
 
-	// Check if there is a headless service for the cluster
-
 	desiredEndpoints := newEndpointsForAdditionalSeeds(dc)
 
 	createNeeded := false
--- a/operator/pkg/reconciliation/reconcile_endpoints.go
+++ b/operator/pkg/reconciliation/reconcile_endpoints.go
@@ -36,8 +36,6 @@ func (rc *ReconciliationContext) CreateEndpointsForAdditionalSeedService() resul
 
 	rc.Recorder.Eventf(rc.Datacenter, "Normal", "CreatedResource", "Created endpoints %s", endpoints.Name)
 
-	// at this point we had previously been saying this reconcile call was over, we're done
-	// but that seems wrong, we should just continue on to the next resources
 	return result.Continue()
 }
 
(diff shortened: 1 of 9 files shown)


tests/additional_seeds/additional_seeds_suite_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+			ns.WaitForOperatorReady()
+
+			step = "creating a datacenter resource with 2 racks/4 nodes and 2 additional seeds"
+			k = kubectl.ApplyFiles(dcYaml)
+			ns.ExecAndLog(step, k)
+
+			ns.WaitForDatacenterReady(dcName)
+
+			checkSeedConstraints()
```
respringer, line 277: I started with a copy of the existing seed integration test.  We can remove these extra validation checks if we want.

tests/testdata/additional-seeds-two-rack-four-node-dc.yaml
```diff
@@ -8,7 +8,8 @@ spec:
   serverVersion: "6.8.1"
```
jimdickinson, line 8: did you mean to `git mv` the old file, or to `cp && git add`?

operator/pkg/reconciliation/reconcile_endpoints.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if err := client.Create(rc.Ctx, endpoints); err != nil {
+		logger.Error(err, "Could not create endpoints for additional seed service")
+
+		return result.Error(err)
+	}
+
+	rc.Recorder.Eventf(rc.Datacenter, "Normal", "CreatedResource", "Created endpoints %s", endpoints.Name)
+
+	// at this point we had previously been saying this reconcile call was over, we're done
```
jimdickinson, line 39: should probably clean up this comment? was more relevant in some old code

tests/testdata/additional-seeds-two-rack-four-node-dc.yaml
respringer: oops.  Will fix.

respringer: Ok, that was a good catch.  I just added a copy from master.  Once this is merged to master, the older file history should win because of how git tracks content.

jimdickinson: squash and merge ftw :)

operator/pkg/reconciliation/reconcile_endpoints.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (rc *ReconciliationContext) CheckAdditionalSeedEndpoints() result.ReconcileResult {
+	// unpacking
+	logger := rc.ReqLogger
+	dc := rc.Datacenter
+	client := rc.Client
+
+	logger.Info("reconcile_endpoints::CheckAdditionalSeedEndpoints")
+
+	desiredEndpoints := newEndpointsForAdditionalSeeds(dc)
```
jimdickinson, line 50: what do you think about adding a guard here that returns result.Continue() early if len(dc.Spec.AdditionalSeeds) == 0?

PR #177, opened 2020-07-28
Ko 461 nodeport service v2

respringer:
A NodePort service may be requested by setting the following fields:

```
  networking:
    nodePort:
      cql: 30001
      broadcast: 30002
 ```

The SSL versions of the ports may be requested:

```
  networking:
    nodePort:
      cqlSSL: 30010
      broadcastSSL: 30020
```
   
If any of the nodePort fields have been configured then a NodePort service will be created that routes from the specified external port to the identically numbered internal port.  Cassandra will be configured to listen on the specified ports.  Additionally, the standard service will no longer define a "native" port.

### CODE REVIEW

File: docs/user/README.md [hunk 1]
```diff
@@ -363,6 +363,24 @@ spec:
...
+    nodePort:
+      cql: 30001
+      broadcast: 30002
+ 
+The SSL versions of the ports may be requested:
+
+  networking:
+    nodePort:
+      cqlSql: 30010
+      broadcastSql: 30020
```
jimdickinson, line 380:
```suggestion
      cqlSsl: 30010
      broadcastSsl: 30020
```

jimdickinson: and actually maybe SSL should be in caps, because golang idioms

mage/ginkgo/lib.go
```diff
@@ -369,3 +367,27 @@ func (ns NsWrapper) HelmInstall(chartPath string) {
 	err := helm_util.Install(chartPath, "cass-operator", ns.Namespace, overrides)
 	mageutil.PanicOnError(err)
 }
+
+// Note that the actual value will be cast to a string before the comparison with the expectedValue
+func (ns NsWrapper) ExpectKeyValue(m map[string]interface{}, key string, expectedValue string) {
```
respringer, line 372: ExpectKeyValues(), which takes a whole map, and the helper function ExpectKeyValue() are a first-pass at a generic way to validate key/values.  Right now everything is cast to a string.  I added a helper for dealing with Float64 values because the Ports from the service were coming back as Float64s for some reason.  I am open to suggestion for improvements, but we may want to handle them in a follow-up ticket.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -569,21 +606,61 @@ func (dc *CassandraDatacenter) GetConfigAsJSON() (string, error) {
...
+// Gets the defined broadcast/intranode port for NodePort.
+// 0 will be returned if NodePort is not configured.
+// The SSL port will be returned if it is defined,
+// otherwise the normal broadcast port will be used.
+func (dc *CassandraDatacenter) GetNodePortBroadcastPort() int {
+	if !dc.IsNodePortEnabled() {
+		return 0
+	}
+
+	if dc.Spec.Networking.NodePort.BroadcastSsl != 0 {
```
respringer, line 636: The port fallback logic is a bit squirrelly and went through multiple designs via slack/standup conversations.  What I have landed on is nodeport ssl ports take first priority, then nodeport non-ssl ports, then the defaults of 9042 for native and 7000 for broadcast.  If the nodeport native/cql port is defined, then the native port will be omitted from the normal service.

docs/user/README.md
respringer: doh!

operator/pkg/reconciliation/constructor.go
```diff
@@ -33,15 +33,20 @@ func newServiceForCassandraDatacenter(dc *api.CassandraDatacenter) *corev1.Servi
...
-		{
-			Name: "native", Port: 9042, TargetPort: intstr.FromInt(9042),
-		},
 		{
 			Name: "mgmt-api", Port: 8080, TargetPort: intstr.FromInt(8080),
 		},
 	}
 
+	// If NodePort is enabled, it will take control of the native port
+	if !dc.IsNodePortEnabled() {
```
jimdickinson, line 42: why disable this when node port is configured vs just put the port number here? I think taking it out might break some folks with in cluster clients

operator/pkg/reconciliation/constructor.go
```diff
@@ -122,6 +127,39 @@ func newEndpointsForAdditionalSeeds(dc *api.CassandraDatacenter) *corev1.Endpoin
...
+	service.Spec.Ports = []corev1.ServicePort{
+		// Note: Port Names cannot be more than 15 characters
+		{
+			Name:       "broadcast",
+			Port:       int32(broadcastPort),
+			NodePort:   int32(broadcastPort),
+			TargetPort: intstr.FromInt(broadcastPort),
+		},
+		{
+			Name:       "native",
```
jimdickinson, line 153: @JeremiahDJordan do you think cql is a better name here? what about broadcast vs. internode?

docs/user/README.md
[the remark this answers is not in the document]
respringer: That was a typo.  I pushed a fix.

mage/ginkgo/lib.go
```diff
@@ -369,3 +367,27 @@ func (ns NsWrapper) HelmInstall(chartPath string) {
...
+	actualValue, ok := m[key].(string)
```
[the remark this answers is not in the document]
respringer: This tries to interpret the value as a string.  If that was possible, then ok is set to true.  If that was not possible, for instance, the value was a float64, then ok is set to false.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -569,21 +606,61 @@ func (dc *CassandraDatacenter) GetConfigAsJSON() (string, error) {
...
+	} else {
+		return 7000
+	}
+}
+
 // GetContainerPorts will return the container ports for the pods in a statefulset based on the provided config
 func (dc *CassandraDatacenter) GetContainerPorts() ([]corev1.ContainerPort, error) {
+
+	cqlPort := 9042
+	broadcastPort := 7000
```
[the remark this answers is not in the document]
respringer: Good idea, I will do that.

johntrimble:
By the way, the PR looks great!

operator/pkg/reconciliation/constructor.go
respringer: I will put it back.

operator/pkg/reconciliation/constructor.go
JeremiahDJordan: Broadcast is wrong. I would call it internode.  Both internode and native protocol (cql) ports have broadcast settings.

PR #182, opened 2020-07-31
Ko 474 helm chart cluster wide install

respringer:
This provides a clusterWideInstall boolean that should install the cass-operator in such a way that it can maintain cassandraDatacenters in all namespaces in the kubernetes cluster.


README.md
```diff
@@ -164,11 +164,12 @@ helm install --namespace=my-custom-namespace cass-operator ./charts/cass-operato
...
 
 ```yaml
+clusterWideInstall: false
 serviceAccountName: cass-operator
+serviceAccountRoleName: cass-operator-service-account
+serviceAccountRoleBindingName: cass-operator-service-account
 clusterRoleName: cass-operator-cluster-role
 clusterRoleBindingName: cass-operator
-roleName: cass-operator
-roleBindingName: cass-operator
```
jimdickinson, line 171: why change these keys?

README.md
```diff
@@ -177,6 +178,15 @@ imagePullPolicy: IfNotPresent
 
 NOTE: Helm does not install a storage-class for the cassandra pods.
 
+If clusterWideInstall is set to true, then the operator will be able to administer cassandraDatacenters in all namespaces of the kubernetes cluster.  A namespace still must provided because some of the kubernetes resources for the operator require one.
+
+Example:
+
+```console
+kubectl create namespace my-custom-namespace
```
jimdickinson, line 186: the example should be something like cass-operator-system, I think, like the way istio creates istio-system

README.md
respringer: There are two reasons.  

1. because during a cluster-wide install, they will be used to create a clusterrole, not a role, so "roleName" becomes a little weird
2. during a cluster-wide install, the existing clusterRoleName key will still be used as-is, and "roleName" will now create a second clusterRole.  Renaming these keys to serviceAccountRoleName clarifies their purpose and helps people to distinguish them from the other clusterRole key.

johntrimble: Does this have any upgrade impact if a user already had installed cass-operator using helm or will it "just work" as long as they make sure they use the new value names when performing a `helm upgrade`?

README.md
```diff
@@ -177,6 +178,15 @@ imagePullPolicy: IfNotPresent
 
 NOTE: Helm does not install a storage-class for the cassandra pods.
 
+If clusterWideInstall is set to true, then the operator will be able to administer cassandraDatacenters in all namespaces of the kubernetes cluster.  A namespace still must provided because some of the kubernetes resources for the operator require one.
```
johntrimble, line 181:
```suggestion
If clusterWideInstall is set to true, then the operator will be able to administer `CassandraDatacenter`s in all namespaces of the kubernetes cluster.  A namespace must still be provided because some of the kubernetes resources for the operator require one.
```

README.md
respringer: I will manually that and see what happens

README.md
```diff
@@ -157,18 +157,19 @@ cluster1-superuser@cqlsh> select * from system.peers;
...
 
 ```yaml
+clusterWideInstall: false
 serviceAccountName: cass-operator
-clusterRoleName: cass-operator-cluster-role
-clusterRoleBindingName: cass-operator
-roleName: cass-operator
-roleBindingName: cass-operator
+serviceAccountRoleName: cass-operator-service-account
+serviceAccountRoleBindingName: cass-operator-service-account
```
jimdickinson, line 170:
```suggestion
roleName: cass-operator
roleBindingName: cass-operator
```

and then we document that these will be used for ClusterRole and ClusterRoleBinding resources when the boolean is flipped to true - how does that sound?

PR #189, opened 2020-08-09
Use the additional seeds service in the config

jimdickinson:
We want to avoid pushing the IPs down into the config, because it will cause extra rolling upgrades.

### DIFF
--- a/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -547,7 +547,9 @@ func (dc *CassandraDatacenter) GetConfigAsJSON() (string, error) {
 	// resolve to the seed nodes. This obviates the need to update the
 	// cassandra.yaml whenever the seed nodes change.
 	seeds := []string{dc.GetSeedServiceName()}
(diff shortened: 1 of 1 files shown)


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -547,7 +547,9 @@ func (dc *CassandraDatacenter) GetConfigAsJSON() (string, error) {
 	// resolve to the seed nodes. This obviates the need to update the
 	// cassandra.yaml whenever the seed nodes change.
 	seeds := []string{dc.GetSeedServiceName()}
-	seeds = append(seeds, dc.Spec.AdditionalSeeds...)
+	if len(dc.Spec.AdditionalSeeds) > 0 {
+		seeds = append(seeds, dc.GetAdditionalSeedsServiceName())
+	}
```
johntrimble, line 552: Err, were we getting `nil` in the seed list or something here?

jimdickinson: no, we were pushing the IPs down into config, which the change to add a Service and Endpoints resources was there to avoid

Issue #174, opened 2020-07-23
Is it possible to use a single operator for multiple cassandra-datacenters across different namespaces?


rltvty:
Can you comment more about the downsides of running one operator per cluster, instead of per-namespace?

In my environment, we will have many independent CassandraDatacenters each in their own kubernetes namespace. Currently, we have 6, 3-node CassandraDatacenters, but could easily scale to 50 or 100 independent CassandraDatacenters in the future, all inside a single kubernetes cluster.

It would be much easier if we could have a single deployment of the operator, maybe with several pods, to manage all the CassandraDatacenters, instead of deploying an operator per namespace.

Thanks for the info.

eldondevat:
Hey rltvty, this is definitely a use case we want to support. We're still building out the testing infrastructure to support this strategy at scale. In this scenario, are the CassandraDatacenters each living alone in a namespace, or do their applications deploy in that namespace as well? If the former, how do you plan on exposing those as services?

rltvty:
@eldondevat The applications using the CassandraDataCenters live in the same namespace.  

Ideally, nothing outside the namespace would communicate with the CassandraDataCenter, other than the shared cassandra-operator.

jimdickinson:
I think the downsides are more in theory than in practice, but just not enough testing has been done for this use case to answer definitively. We can get some small tickets on the roadmap though that should improve your experience. Here are some I can think of...

* Helm install option for cluster-wide / single-namespace
* Better logging to count the reconciles done per CassDC, not just assigning each reconcile loop iteration a uuid.
* Generic back-off / cooldown for clusters that just completed a major milestone in reconciling. We have one for starting a Cassandra node, but could use one for updating the statefulset, or completing reconciling with no more work to do. In practice I've seen the eventual consistency cause extra reconcile spam after we just completed getting a DC up and running.

Some more questions: One Cassandra DC per k8s namespace? Are you planning on having k8s workers with multiple Cassandra pods on them, or no?

(some relevant context - https://github.com/datastax/cass-operator/issues/115#issuecomment-663192818)

rltvty:
On the other questions:

Yes, one CassDC per k8s namespace.  And ideally yes we would have k8s workers (nodes) with multiple Cassandra pods on them (from different namespaces).

jimdickinson:
[...]

PR #190, opened 2020-08-09
Add quietPeriod and observedGeneration to the status

jimdickinson:
This PR helps the operator work well when there are many CassandraDatacenters to reconcile, needed for #174. We're adding the concept of a "quiet period" where we can delay reconciling work for a given CassDC when there's unlikely to be anything for the operator to do.
* When we're rolling out a configuration upgrade we can leave some time for pods to be terminated.
* When the CassDC reaches a state where there's no more work to do we can take a few seconds off. (There's usually a few reconcile loops that check the entire state and find nothing to do - this is seen in Cassandra logs where we reload seed configuration several times in a row. This PR mitigates that.)

I tacked on adding observedGeneration here - it helps when writing automation around the Cass Operator.

### DIFF
--- a/operator/pkg/reconciliation/constructor.go
+++ b/operator/pkg/reconciliation/constructor.go
@@ -7,8 +7,9 @@ package reconciliation
 
 import (
 	"fmt"
-	"k8s.io/api/batch/v1"
 	"os"
+
+	v1 "k8s.io/api/batch/v1"
 
 	api "github.com/datastax/cass-operator/operator/pkg/apis/cassandra/v1beta1"
 	"github.com/datastax/cass-operator/operator/pkg/httphelper"
@@ -676,7 +677,7 @@ func buildInitReaperSchemaJob(dc *api.CassandraDatacenter) *v1.Job {
 	}
 }
 
-func addVolumes(dc *api.CassandraDatacenter,baseTemplate *corev1.PodTemplateSpec) []corev1.Volume {
+func addVolumes(dc *api.CassandraDatacenter, baseTemplate *corev1.PodTemplateSpec) []corev1.Volume {
 	vServerConfig := corev1.Volume{}
 	vServerConfig.Name = "server-config"
 	vServerConfig.VolumeSource = corev1.VolumeSource{
@@ -693,9 +694,8 @@ func addVolumes(dc *api.CassandraDatacenter,baseTemplate *corev1.PodTemplateSpec
 	vServerEncryption.Name = "encryption-cred-storage"
 	vServerEncryption.VolumeSource = corev1.VolumeSource{
 		Secret: &corev1.SecretVolumeSource{
-			SecretName:fmt.Sprintf("%s-keystore", dc.Name)},
+			SecretName: fmt.Sprintf("%s-keystore", dc.Name)},
 	}
-
 
 	volumes := []corev1.Volume{vServerConfig, vServerLogs, vServerEncryption}
 	baseTemplate.Spec.Volumes = append(baseTemplate.Spec.Volumes, volumes...)
--- a/operator/pkg/reconciliation/constructor.go
+++ b/operator/pkg/reconciliation/constructor.go
@@ -368,6 +368,7 @@ func setOperatorProgressStatus(rc *ReconciliationContext, newState api.ProgressS
 
 	patch := client.MergeFrom(rc.Datacenter.DeepCopy())
 	rc.Datacenter.Status.CassandraOperatorProgress = newState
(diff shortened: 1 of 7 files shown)


operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -264,9 +264,15 @@ func (rc *ReconciliationContext) CheckRackPodTemplate() result.ReconcileResult {
 				return result.Error(err)
 			}
 
+			if err := rc.enableQuietPeriod(20); err != nil {
```
johntrimble, line 267: Is there any reason to not just pause here for 20 seconds, as opposed to doing nothing on subsequent reconcile calls for 20 seconds?

operator/pkg/reconciliation/constructor.go
```diff
@@ -367,6 +368,9 @@ func setOperatorProgressStatus(rc *ReconciliationContext, newState api.ProgressS
 
 	patch := client.MergeFrom(rc.Datacenter.DeepCopy())
 	rc.Datacenter.Status.CassandraOperatorProgress = newState
+	if newState == api.ProgressReady {
+		rc.Datacenter.Status.ObservedGeneration = rc.Datacenter.Generation
```
johntrimble, line 372: 💯

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -275,7 +281,8 @@ func (rc *ReconciliationContext) CheckRackPodTemplate() result.ReconcileResult {
 			// because there's an upgrade in progress
 
 			status := statefulSet.Status
-			if status.Replicas != status.ReadyReplicas ||
+			if statefulSet.Generation != status.ObservedGeneration ||
```
johntrimble, line 284: :100: Thanks for fixing this!

operator/pkg/reconciliation/constructor.go
johntrimble: Err, actually, I don't think this is quite right. The ObservedGeneration and the actual generation being equal doesn't mean the resource is ready. All it means is that the operator has at least seen the generation in question. For example, when the ObservedGeneration for a statefulset equals the actual generation, that does not mean the statefulset is up and running, it just means the controller has had a chance to respond to the change of that resource.

jimdickinson: right - we've decided it's ready and then we set the ObservedGeneration - this is to provide something better than what we have right now, where you don't know if the CassDC is ready because the operator hasn't started work on changing it, or if it's all done in under a couple milliseconds

Someone working on automation around Cass Operator chatted me the other day...

> Matching generation would mean like "the requested spec has been successfully fully propagated at least once"

johntrimble: I think the readiness of the cluster and the value of the observed generation are orthogonal. If you want to know if the requested spec has been fully propagated, you would check the `DatacenterReady` condition _and_ check the observed generation. The status tells you what the current state is, the observed generation tells you if the status is up-to-date. No one should rely exclusively on the observed generation to know if the current spec is fully propagated. That's not how statefulsets work for example. The observed generation equalling the current generation doesn't really have anything to do with readiness, it just means the operator has seen the current generation.

jimdickinson: That's right - check both - sorry I didn't get that exact phrase out in my last reply 😇

operator/pkg/reconciliation/reconcile_racks.go
jimdickinson: Yes - for #174 we want one Cass Operator to handle more CassDCs gracefully, so blocking in the middle of the reconcile loop for any of them is not great.

operator/pkg/reconciliation/constructor.go
johntrimble: I guess I'm still confused as to why we are only doing this when `ProgressReady`. We should do this regardless of the state at the end of reconciliation (unless we are in a quiet period). The value of `CassandraOperatorProgress` does not seem relevant here.

operator/pkg/reconciliation/reconcile_racks.go
johntrimble: That makes sense. :+1:

operator/pkg/reconciliation/constructor.go
jimdickinson: We haven't fully observed and pushed down the change until all our writes to k8s resources are done, right?

PR #195, opened 2020-08-12
Toleration support v2

respringer:
This PR is purely for review, a follow-up PR will contain these changes and additional logic.

This adds a watch for a couple of vmware Node taints.

Note that the watch requires a new clusterrole and clusterrolebinding to be added to the helm chart for the cass-operator.

If the taints are found, the corresponding pods on those Nodes will be drained and deleted.  Note that the pod will probably not automatically be re-assigned to a new Node because the existing PersistentVolume has node affinity for the tainted node.  That issue will be addressed in a follow-up ticket.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,8 @@ The following Helm default values may be overridden:
 ```yaml
 clusterWideInstall: false
 serviceAccountName: cass-operator
+clusterRoleName: cass-operator-cr
+clusterRoleBindingName: cass-operator-crb
 roleName: cass-operator
 roleBindingName: cass-operator
 webhookClusterRoleName: cass-operator-webhook
--- a/operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
+++ b/operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
@@ -4,8 +4,9 @@
 package cassandradatacenter
 
 import (
+	"context"
 	"fmt"
-	
+
 	api "github.com/datastax/cass-operator/operator/pkg/apis/cassandra/v1beta1"
 
 	"github.com/datastax/cass-operator/operator/pkg/oplabels"
@@ -18,6 +19,11 @@ import (
 
 	"github.com/datastax/cass-operator/operator/pkg/reconciliation"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	types "k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -166,6 +166,8 @@ The following Helm default values may be overridden:
 ```yaml
 clusterWideInstall: false
 serviceAccountName: cass-operator
+clusterRoleName: cass-operator-cr
```
respringer, line 169: The "cass-operator" name for a cluster-role will already be taken if the helm chart has clusterWideInstall=true.  Therefore I added the -cr and -crb suffixes.

tests/tolerations/tolerations_suite_test.go
```diff
@@ -0,0 +1,95 @@
...
+			step = fmt.Sprintf("tainting node: %s", node1Name)
+			k = kubectl.Taint(
+				node1Name,
+				"node.vmware.com/drain",
+				"planned-downtime",
+				"NoSchedule")
+			ns.ExecAndLog(step, k)
+
+			time.Sleep(1 * time.Minute)
+
```
respringer, line 81: In the future, I will add logic to ensure that the pod comes back up on a new node.  Currently it becomes stuck in Pending after being deleted.

operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
```diff
@@ -112,6 +119,72 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	// Setup watches for Nodes to check for taints being added
+
+	mapFn := handler.ToRequestsFunc(
+		func(a handler.MapObject) []reconcile.Request {
+			log.Info("Node Watch called")
+			requests := []reconcile.Request{}
+
+			nodeName := a.Object.(*corev1.Node).Name
+
+			c := mgr.GetClient()
```
johntrimble, line 131: So we've generally been quite cautious to avoid hitting the k8s apis in these handlers, in fact, we've gone so far as to maintain back references in metadata to lookup the relevant resources to enqueue requests for so that we don't have to hit the k8s APIs (the elastic operator does something similar). This function will execute every time the most mundane of changes happens to a node. Do we know of any other operator out there that uses the client from one of these EventHandlers?

respringer: Why are we concerned about using the manager client from inside of a watch handler function?  

The watches are running in the same process ( modulo threading/go-routines ) as the controller itself.  Calling the k8s api from a watcher should be virtually identical as calling the k8s api from the reconcile loop, right?

I am not opposed to maintaining a list of relevant pods in the node metadata, but I am not sure what the concern is.

johntrimble: So this might be okay, it's been a while since I've crawled through the controller runtime code. When I've looked into this previously I found that the elasitic operator went out of its way to avoid using the client here, and I've not seen an example of another operator doing this. So if you feel you've done your homework and believe this okay, I'm on board.

One problem you do seem to have is that if you encounter an error for any reason when querying resources, you don't really have a way to try again. So if you fail to list all the pods (due to some transient error, like maybe etcd is under load) then you will effectively lose the event (and you might not get a subsequent one). In reconcile you have the advantage of being able to try again if something goes wrong, but you cannot do that here, at least not as far as I know.

respringer: I am not sure exactly how to deal with an error in the watcher mapFunction, except to log it.  The worst case scenario is that the cass-operator doesn't notice that a node has the new taint, and the node is subsequently taken down suddenly, taking the pod and possibly a pv with it.  In that case, we are relying on the normal redundancy features of cassandra to handle the loss of the node, which we might need anyway if we can't even get a list of pods from etcd.

If you have a better strategy idea, I'd be happy to give it a go.

johntrimble: So errors in the mapping function lead to lost events. Errors during reconciliation lead to a retry. If you build the mapping of nodes to pods (or nodes to datacenters) during reconciliation, you will not have to worry about lost events. Then, in the event handler, you can look at the mapping to find the relevant datacenters without having to do any IO (which might return an error, especially if it is over the network). The dynamic watches does this, but you would probably want a simpler in memory map like how elastic does it.

Issue #167, opened 2020-07-15
Validate all ports are defined within the StatefulSet


bradfordcp:
Currently many ports are defined on the STS, but not all of them. This includes both Cassandra ports (JMX) and those leveraged by DSE. In most circumstances this is not a problem when running everything within the k8s network. Unfortunately this information is used by K8s when using host network deployments IE external ports are exposed and forwarded to the STS. If the STS is not aware of a port then it will not be forwarded. Please see the reference list below:

### Apache Cassandra

* 7000
* 7001
* 7199
* 9042
* 9103
* 9160
* 9142

### DataStax Enterprise

* 3030
* 5598
* 5599
* 7000
* 7001
* 7077
* 7080
* 7081
* 7199
* 8090
* 8182
* 8609
* 8983
* 9042
* 9077
* 9091
* 9103
* 9142
* 9160
* 9999
* 10000
* 18080

jeromatron:
So will this open all the specified ports for every workload or will it be dependent upon the workload that is enabled or is it simply configurable?

jtyl2017:
> So will this open all the specified ports for every workload or will it be dependent upon the workload that is enabled or is it simply configurable?

e.g. if graph advanced workload is being deployed, only graph related ports (+ standard cassandra ports) will be added to statefulset

jeromatron:
So what is currently not in the operator even in 1.4 is the DSE workload specific ports like for Search. It would be nice if those ports could be opened depending on what workload is enabled.

PR #200, opened 2020-08-17
More container ports

jimdickinson:
Working on #167


jimdickinson:
> Everything looks good here. I am curious if we want to get rid of some of the magic numbers regarding ports and define some constants in a common location. It's not a blocker, more of a question.

Yeah, we'll clean this up in the next PR.

PR #204, opened 2020-08-19
Toleration support v2

respringer:
This PR adds support for detecting two vmware node taints.  If either one is detected, then any corresponding pods on those nodes will have their Cassandra nodes drained, and their PersistentVolumeClaims and pods deleted.  The statefulset will then recreate the pods and pvcs.

taint name: node.vmware.com/drain
taint key: planned-downtime or drain
taint effect: NoSchedule

Note that a known kubernetes bug will cause the statefulset to first recreate the pod in an eternal Pending status because the statefulset does not properly recreate the pvc.  The operator will detect this condition and delete that stuck pod.  At this point the statefulset will then properly recreate a pvc and a functional pod.

See https://github.com/kubernetes/kubernetes/issues/89910


operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1563,6 +1586,34 @@ func (rc *ReconciliationContext) findStartedNotReadyNodes() (bool, error) {
 	return false, nil
 }
 
+func (rc *ReconciliationContext) copyPodCredentials(pod *corev1.Pod, jksBlob []byte) error {
```
respringer, line 1589: This function was brought in from a merge of master into my branch.  Was this a failed merge conflict resolution?  Should I delete this function? @johntrimble ?

tests/testdata/tolerations-dc.yaml
```diff
@@ -0,0 +1,32 @@
...
+          requests:
+            storage: 1Gi
+  racks:
+    - name: r1
+  config:
+    jvm-server-options:
+      initial_heap_size: "800m"
+      max_heap_size: "800m"
+      additional-jvm-opts:
+        - "-Dcassandra.system_distributed_replication_dc_names=dc1"
```
respringer, line 27: All three of these settings are necessary to fully bring up the Cassandra service in the recreated pod.

operator/pkg/reconciliation/reconcile_racks.go
johntrimble: I don't know what that is. It looks like something from one of Eldon's PRs.

operator/pkg/reconciliation/check_nodes.go
```diff
@@ -0,0 +1,183 @@
...
+	err := rc.Client.Get(rc.Ctx, types.NamespacedName{Namespace: podNamespace, Name: pvcFullName}, pvc)
+	if err != nil {
+		rc.ReqLogger.Error(err, "error retrieving PersistentVolumeClaim")
+		return nil, err
+	}
+
+	return pvc, nil
+}
+
+func (rc *ReconciliationContext) DeletePvcIgnoreFinalizers(podNamespace string, podName string) (*corev1.PersistentVolumeClaim, error) {
```
johntrimble, line 28: Ugh. What I want to say we should do here is, if there is a non-blocking way to delete the PVC, just do that and then delete the Pod. When the terminates, the finalizer `kubernetes.io/pvc-protection` should be removed automagically and deletion of the PVC should then complete. This is making me wonder if we should _just_ delete the pod when we see a node taint, and then handle Pods that cannot be scheduled due to the node affinity of their PVC separately (i.e. don't do a replace, just delete the pod, if we see a stuck pod due to a PVC issue, then delete the PVC and do a replace).  We may end up just leaving it this way for now though...

johntrimble: Err... also, deleting the PVC out from under the pod might cause the pod to be unable to terminate cleanly... not sure if that matters in the case of Cassandra...

operator/pkg/reconciliation/reconcile_racks.go
respringer: I will remove it.

operator/pkg/reconciliation/handler.go
```diff
@@ -27,6 +30,56 @@ import (
...
 
+// key: Node.Name, value: CassandraDatacenter.Name
+var nodeToDcName = make(map[string]string)
+
+// key: Node.Name, value: CassandraDatacenter.Namespace
+var nodeToDcNamespace = make(map[string]string)
+
+// Get the dcName and dcNamespace for a node
+func DatacenterForNode(nodeName string) (string, string) {
+	dcName, ok := nodeToDcName[nodeName]
```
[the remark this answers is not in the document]
respringer: I spent some time thinking about that, but I figured the mutex wouldn't be necessary because only one thread would ever be writing to this thing at a time.  I typically use mutex's to avoid multiple writers.  The reader is doing a check for key existence, which should either return true or false.  If a key is found, then a reconcile request is triggered, if not, nothing happens.

If a key hasn't been properly removed, then we suffer a needless reconcile call.  
If a key hasn't been properly added, then we miss reporting a node change event.  However, I find this situation highly unlikely, because we don't actually care about node events until our pods are being created, in which case we have probably gone through a few rounds of reconcile loops that properly updated this map.

In any case, I don't mind adding a mutex to be extra safe.  I just didn't see a pressing need for it.

operator/pkg/reconciliation/handler.go
```diff
@@ -27,6 +30,56 @@ import (
...
+	if ok {
+		return dcName, nodeToDcNamespace[nodeName]
```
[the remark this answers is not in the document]
respringer: Both of those are good catches.  I'll go fix them.

operator/pkg/reconciliation/handler.go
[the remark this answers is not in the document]
johntrimble: Looks like maps are not safe for concurrent read/write:

https://stackoverflow.com/questions/36167200/how-safe-are-golang-maps-for-concurrent-read-write-operations

If I understand the comments there correctly, it used to be okay to have a concurrent read with a write, so long as there were not concurrent writers, but that is no longer the case. I still don't know if the read and write are actually happening in separate threads though.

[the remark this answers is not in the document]
respringer: TIL.  Ok, i'll wrap it in a concurrency primitive.

PR #220, opened 2020-08-25
Add test for infinite reconcile

johntrimble:


mage/ginkgo/lib.go
```diff
@@ -400,3 +400,23 @@ func (ns NsWrapper) ExpectKeyValues(actual map[string]interface{}, expected map[
...
+	time.Sleep(1*time.Minute)
+
+	json := `jsonpath={.metadata.resourceVersion}`
+	k := kubectl.Get("CassandraDatacenter", dcName).
+		FormatOutput(json)
+	resourceVersion := ns.OutputPanic(k)
+
+	time.Sleep(1*time.Minute)
+
+	json = `jsonpath={.metadata.resourceVersion}`
```
jimdickinson, line 415: is this enough to check? I thought I've seen this stay the same...

I know we've got somewhere to get the logs from the operator - what I always do is grab them and compare the `tail`

johntrimble: This will catch issues where either the status or the resource itself has been updated, which I think is all the cases we've encountered thus far. It is possible that the operator could change some _other_ resource, which triggers reconciliation, which changes the other resource again, which triggers reconciliation... We could look at the loop identifier in the log. Do you want me to fix that now or is this good enough for the time being.

jimdickinson: sounds good to me

tests/oss_test_all_the_things/oss_test_all_the_things_suite_test.go
```diff
@@ -101,6 +101,8 @@ var _ = Describe(testName, func() {
 
 			ns.WaitForDatacenterReady(dcName)
 
+			ns.ExpectDoneReconciling(dcName)
```
jimdickinson, line 104: thank you for plugging into this test too!

Issue #211, opened 2020-08-24
Reaper does not work when Cassandra authentication is used


jsanda:
Consider the following example `CassandraDatacenter` manifest:

```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: cluster1
  serverType: cassandra
  serverVersion: "3.11.6"
  managementApiAuth:
    insecure: {}
  size: 3
  storageConfig:
    cassandraDataVolumeClaimSpec:
      storageClassName: server-storage
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
  reaper:
    enabled: true
  config:    
    cassandra-yaml:
      authenticator: org.apache.cassandra.auth.PasswordAuthenticator
      authorizer: org.apache.cassandra.auth.CassandraAuthorizer
      role_manager: org.apache.cassandra.auth.CassandraRoleManager
```

Cass Operator will create a C* cluster with authentication enabled. Reaper talks to Cassandra using both JMX and CQL. The CQL calls will fail because no credentials are passed to Reaper.

It might be good for `.spec.reaper` to allow you to specify a C* user/role for Reaper to use. I think it is sufficient for now though for the operator to auto generate a role for Reaper.

jsanda:
This may require some more effort and changes that I initially anticipated. For its readiness probe, Reaper attempts to connect to its storage backend, which is Cassandra. In order for the readiness check to succeed, the `reaper_db` schema needs to exist. Cass Operator deploys a k8s job to create the keyspace. The Casandra pods which includes both Cassandra and Reaper are deployed before the job. There may be some Reaper container restarts until the job has a chance to run.

When authentication is enabled, the operator needs to create a C* role before the schema initialization job can run. This makes things more complicated. In the reconciliation loop today the schema initialization job does not run the statefulsets are ready. The statefulsets however cannot reach the ready state until both the Cassandra and Reaper containers reach the ready state. Chicken meet egg.

I should also point out that in my prototyping I am creating a separate C* role for Reaper. I am trying to piggyback off the existing user creation code. Specifically, I have modified the `GetUsers` function in `reconcile_racks.go` to add a user for Reaper when Reaper and authentication are enabled. `GetUsers` is called by `CreateUsers`. 

[...]

PR #222, opened 2020-08-25
Use C* authentication with Reaper

jsanda:
This is for #211.


operator/pkg/reconciliation/reconcile_reaper.go
```diff
@@ -106,7 +145,16 @@ func (rc *ReconciliationContext) CheckReaperSchemaInitialized() result.Reconcile
...
+		patch := client.MergeFrom(rc.Datacenter.DeepCopy())
+		rc.Datacenter.Status.ReaperStatus.SchemaInitialized = true
+		if err = rc.Client.Status().Patch(rc.Ctx, rc.Datacenter, patch); err != nil {
+			rc.ReqLogger.Error(err, "error updating the reaper status")
+			return result.Error(err)
+		}
+		// Requeue with a delay to give a chance for C* schema changes to propagate
+		//
+		// TODO Should the delay be adjusted based on the C* cluster size?
+		return result.RequeueSoon(5)
```
johntrimble, line 157: Is there any way to ask Cassandra if the changes are done propagating? Like a query we can run or something?

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -553,6 +559,32 @@ func (dc *CassandraDatacenter) GetSuperuserSecretNamespacedName() types.Namespac
...
+
+func (dc *CassandraDatacenter) IsAuthenticationEnabled() (bool, error) {
+	b, err := dc.Spec.Config.MarshalJSON()
+	if err != nil {
+		return false, err
+	}
+
+	s := string(b)
+
+	return strings.Contains(s, "PasswordAuthenticator"), nil
```
johntrimble, line 585: Hmmm... so this config you have here isn't the full story. This will get handed to the config builder that will actually generate the cassandra.yaml. In the past, the defaults for the config builder have sometimes varied from that of the actual defaults (as we do with the SeedProvider impl). Right now, we use the "AllowAllAuthenticator" default in configbuilder, but that's the sort of thing I could see changing. Is there any reason we cannot just assume authentication is enabled?  I think we always create a superuser with credentials irrespective of the authentication setting.

johntrimble: Also, if you look at `json.RawMessage` you'll see it's just a `[]byte`:

```
type RawMessage []byte
```

So, if you are going to cheat and not actually parse the JSON, you can just do:

```
return strings.Contains(string(dc.Spec.Config), "PasswordAuthenticator")
```

Then you don't have to return an error from the function either. If the JSON is ill-formed, it will blow up elsewhere.

operator/pkg/reconciliation/secrets.go
```diff
@@ -50,6 +50,12 @@ func generateUtf8Password() (string, error) {
 }
 
 func buildDefaultSuperuserSecret(dc *api.CassandraDatacenter) (*corev1.Secret, error) {
+	//if dc.ShouldGenerateSuperuserSecret() {
+	//	return buildUserSecret(dc.GetSuperuserSecretNamespacedName(), dc.Spec.ClusterName + "-superuser")
+	//
+	//}
+	//
+	//return nil, nil
```
johntrimble, line 58: From debugging?

tests/testdata/oss-three-node-with-auth-dc.yaml
```diff
@@ -0,0 +1,32 @@
...
+      resources:
+        requests:
+          storage: 1Gi
+  racks:
+    - name: r1
+    - name: r2
+    - name: r3
+  config:
+    cassandra-yaml:
+      authenticator: org.apache.cassandra.auth.PasswordAuthenticator
```
johntrimble, line 27: If you tell reaper to use auth and give it the credentials for an actual user, but the PasswordAuthenticator is not actually enabled, will it still work?

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -6,13 +6,13 @@ package v1beta1
...
 	"fmt"
-	"os"
-
 	"github.com/Jeffail/gabs"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"os"
+	"strings"
```
jimdickinson, line 15: nit - stdlib imports should be the first group

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
jimdickinson: I agree with assuming Authentication is enabled - because will all of this work if they toggle it on after setup?

operator/pkg/reconciliation/reconcile_reaper.go
jsanda: There is. C* driver provide an API to check for schema agreement. With the Java driver for example, it is just a single method call, `checkSchemaAgreement()`, that returns a boolean. Internally, the driver executes the following queries:

```
SELECT peer, rpc_address, schema_version FROM system.peers
```
and

```
SELECT schema_version FROM system.local WHERE key='local'
```
It checks that `schema_version` is the same across all rows.

It would be easy to just add the `checkSchemaAgreement` call to the schema init job; however, I plan to get rid of the schema init job now that the management api supports creating keyspaces. Given that I will implement a `checkSchemaAgreement` function in the operator.

jsanda:
So.... This has turned into quite a bit of work which includes changes in Reaper. Please see https://github.com/thelastpickle/cassandra-reaper/issues/956 for some additional context. 

There are two things I want to point out with the latest commits I pushed. First, I have temporarily added a second job for applying Reaper's schema. The initial, original job runs a python script that creates the keyspace. The second job runs a modified version of Reaper that only applies schema updates. I plan to merge all of this into a single image which can then be run in one job.

Secondly, I have added code for configure the replication_factor of the system_auth keyspace. If authentication is enabled and system_auth is using RF = 1, there is a lot of flapping with the `reaper` containers. This is due to the fact that the C* node which has the auth data might not be up yet. This resulted in my debugging a lot of inconsistent test failures. This change is not specific to Reaper though. When auth is enabled, system_auth should never use RF = 1. And even if auth is disabled it doesn't hurt to the RF because the amount of data involved is very small.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -6,16 +6,14 @@ package v1beta1
...
+	"github.com/datastax/cass-operator/operator/pkg/serverconfig"
+	"github.com/datastax/cass-operator/operator/pkg/utils"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-
-	"github.com/datastax/cass-operator/operator/pkg/serverconfig"
-	"github.com/datastax/cass-operator/operator/pkg/utils"
+	"os"
```
jimdickinson, line 16: nit: this import switch should be reverted

operator/pkg/httphelper/client.go
```diff
@@ -236,6 +237,52 @@ func (client *NodeMgmtClient) CallKeyspaceCleanupEndpoint(pod *corev1.Pod, jobs
 	return err
 }
 
+type ReplicationSetting struct {
+	Datacenter string
+	ReplicationFactor int
+}
+
+func (client *NodeMgmtClient) CallAlterKeyspaceEndpoint(pod *corev1.Pod, keyspace string, replicationSettings[]ReplicationSetting) error {
```
jimdickinson, line 245:
```suggestion
func (client *NodeMgmtClient) CallAlterKeyspaceEndpoint(pod *corev1.Pod, keyspace string, replicationSettings []ReplicationSetting) error {
```

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -825,6 +845,36 @@ func (rc *ReconciliationContext) CreateUsers() result.ReconcileResult {
...
+
+	if err != nil {
+		return result.Error(err)
+	}
+
+	return result.Continue()
+}
+
+func getSystemAuthRF(dc *api.CassandraDatacenter) int {
+	return int(math.Min(float64(3), float64(dc.Spec.Size)))
```
johntrimble, line 875: I wonder if we should use the number of racks, instead of the number of nodes, in this `math.Min()` call. In the `isClusterHealthy()` function, we seem to use the rack count as just such a proxy for the RF:

```
func (rc *ReconciliationContext) isClusterHealthy() bool {
	pods := FilterPodListByCassNodeState(rc.clusterPods, stateStarted)

	numRacks := len(rc.Datacenter.GetRacks())
	for _, pod := range pods {
		err := rc.NodeMgmtClient.CallProbeClusterEndpoint(pod, "LOCAL_QUORUM", numRacks)
		if err != nil {
			return false
		}
	}

	return true
}
```

So maybe `return int(math.Min(float64(3), float64(dc..GetRacks())))` instead?

jsanda: @johntrimble what about the scenario though in which we have just a single rack?

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
jsanda: @jimdickinson by any chance are there any project style settings that I can import into GoLand? Or are there style guidelines available so that I can configure GoLand to be consistent?

operator/pkg/reconciliation/reconcile_racks.go
johntrimble: What I'm observing in the code is that we use the number of racks to indicate the intended RF, as in the example above. Is using the number of racks not a suitable proxy for the intended RF? My expectation would be that the only case in which we would not have 3 racks would be the case in which a user is creating a cluster of only 1 node. Is there a reason a user would want to create a cluster of 3+ nodes with only a single rack? 

It's possible the code example I gave above is what needs to change, but I'd like for us to do this in a consistent manner.

jsanda: > Is using the number of racks not a suitable proxy for the intended RF?

Having RF = num_racks is definitely a good practice.

> My expectation would be that the only case in which we would not have 3 racks would be the case in which a user is creating a cluster of only 1 node. Is there a reason a user would want to create a cluster of 3+ nodes with only a single rack?

My understanding is that the operator allows you to specify both the number of nodes and the number of racks. I know that the operator will try to create balanced racks, but the key thing for me here is that there is nothing preventing me from creating a C* cluster with just one rack. If I am only running in a single zone, then a single rack is fine with respect to C* topology.

If we determine the RF of system_auth by the number of racks, then we can wind up with RF = 1 for system_auth even if there are multiple nodes. That would be bad, and not just for Reaper.

jimdickinson:
#307 supersedes this one

PR #223, opened 2020-08-26
Support scaling down a dc/decommissioning nodes

sandoichi:


johntrimble:
Looks like unit tests are failing.

operator/pkg/reconciliation/decommission_node.go
```diff
@@ -0,0 +1,228 @@
...
+func (rc *ReconciliationContext) CheckDecommissioningNodes(epData httphelper.CassMetadataEndpoints) result.ReconcileResult {
+	for _, pod := range rc.dcPods {
+		if pod.Labels[api.CassNodeState] == stateDecommissioning {
+			if !IsDoneDecommissioning(pod, epData) {
+				rc.ReqLogger.Info("Node decommissioning, reconciling again soon")
+			} else {
+				rc.ReqLogger.Info("Node finished decommissioning")
+				rc.ReqLogger.Info("Deleting pod PVCs")
+
+				err := rc.DeletePodPvcs(pod)
```
johntrimble, line 116: So most of the time, PVC have a finalizer on them that prevents them from being deleted while attached to a pod, but it looks like we delete the PVC while the pod is still around. Did that not cause issues in your testing? I think Ryan just recently had that issue. We could find the PVCs after scaling down the statefulset and the pods being terminated, or remove the the finalizers to delete the PVC. Also, I take it that when a pod is decommissioned, the management API continues running so the pod doesn't automatically terminate?

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -2112,17 +2112,6 @@ func (rc *ReconciliationContext) ReconcileAllRacks() (reconcile.Result, error) {
...
-	// TODO until we ignore status updates as it pertains to reconcile
-	// we can't switch in to a quiet period here because it will create
-	// another reconcile iteration with (likely) no work to do
-
-	// if err := rc.enableQuietPeriod(5); err != nil {
-	// 	logger.Error(
-	// 		err,
-	// 		"Error when enabling quiet period")
-	// 	return result.Error(err).Output()
-	// }
```
johntrimble, line 2124: Actually, Jim might want to uncomment this code now that we've fixed it in master.

tests/decom_nodes/decom_nodes_suite_test.go
```diff
@@ -0,0 +1,107 @@
...
+				WithFlag("field-selector", fmt.Sprintf("metadata.name=%s", podWithDecommissionedNode)).
+				FormatOutput(json)
+			ns.WaitForOutputAndLog(step, k, "[]", 30)
+
+			step = "check that the decomm'd pod's PVCs got terminated"
+			json = "jsonpath={.items}"
+			k = kubectl.Get("pvc").
+				WithFlag("field-selector", fmt.Sprintf("metadata.name=%s", podPvcName)).
+				FormatOutput(json)
+			ns.WaitForOutputAndLog(step, k, "[]", 30)
```
johntrimble, line 92: Okay, this is pretty persuasive that things are working as they should be.

operator/pkg/reconciliation/decommission_node.go
johntrimble: Actually, maybe nothing to do here since you have a working test.

sandoichi: Yeah I didn't hit any issues with my testing, but that could definitely be a concern. 

I chose to delete the PVC first because I was worried about a scenario in which we delete the pod first but then somehow fail to delete the PVC, we would never catch it on the next reconcile loop because there would be no pod left to perform a PVC lookup against and it would be orphaned out there until someone caught it.

Although I suppose we could potentially just look up all PVCs that exist and compare them against our current pods and see if there are any anomalies for that case, but we would need to do that check on every single reconcile loop (maybe that's OK?)

johntrimble: To be clear, I'm fine with merging as things are. However, looking at the PVC:

```
  labels:
    app.kubernetes.io/managed-by: cass-operator
    cassandra.datastax.com/cluster: cluster1
    cassandra.datastax.com/datacenter: dc1
    cassandra.datastax.com/rack: r2
  name: server-data-cluster1-dc1-r2-sts-0
``` 

Now I wouldn't trust the "managed-by" label in this case, as in 1.0 the value of this label was wrong and couldn't be fixed for PVCs belonging DCs that already existed, but what you could do is get all PVCs for the DC and use the name to determine if the PVC should be removed. You don't have to do this all the time, just when your condition ScalingDown is true.

I know you have other work related to this, so maybe this is just something to think about for now and potentially address in one of your other tickets.

sandoichi: Ah right, the "scalingDown" condition would still be true even if we failed to delete the pvc during the decomm loop, so we would know to do the check next time. That shouldn't be too bad then.

johntrimble: Actually, I've been thinking about this more, and I think you have the right of it. I was thinking of `kubectl` which will block until the finalizers run, but I don't think that's the case with the go-client? In which case, you have this right. You delete the PVC, the finalizer prevents the PVC from going away, but then you scale down and the statefulset controller removes the pod, which removes the finalizer from the PVC, which allows the PVC to be garbage collected. Sorry for the confusion, what you have should be correct.

You _might_ need to set DeletePropagation (https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#DeletionPropagation) to DeletePropagationBackground, but if it is currently working, then you are probably fine.

operator/example-cassdc-yaml/dse-6.8.x/example-cassdc-minimal.yaml
```diff
@@ -23,3 +23,7 @@ spec:
     jvm-server-options:
       initial_heap_size: "800M"
       max_heap_size: "800M"
+      additional-jvm-opts:
+        # As the database comes up for the first time, set system keyspaces to RF=3
+        - "-Ddse.system_distributed_replication_dc_names=dc1"
+        - "-Ddse.system_distributed_replication_per_dc=3"
```
jimdickinson, line 29: 💯  I noticed after we spoke that this was missing on this example. Can you put it in the Cassandra minimal one too?

tests/testdata/default-three-rack-four-node-dc.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: cassandra.datastax.com/v1beta1
+kind: CassandraDatacenter
+metadata:
+  name: dc1
+spec:
+  clusterName: cluster1
+  serverType: dse
+  serverVersion: "6.8.1"
```
jimdickinson, line 8:
```suggestion
  serverVersion: "6.8.2"
```

operator/pkg/reconciliation/reconcile_racks.go
jimdickinson: Yeah I think we should bring this back, can do it in a separate PR though.

PR #228, opened 2020-08-31
Add support to override the default registry

johntrimble:
This PR adds support for overriding the container registry used to pull required images. It assumes that if a user both specifies an override registry for all images and then explicitly specifies an image from a _different_ registry on a CassandraDatacenter resource that the more specific statement should win (in this case, explicitly state the image to use on the resource wins over the override registry).


mage/k3d/lib.go
```diff
@@ -66,9 +68,43 @@ func createCluster() {
 	mageutil.PanicOnError(err)
 }
 
+func withExponentialBackoff(f func() error, baseWait, timeout time.Duration) error {
```
jimdickinson, line 71: I recommend using https://github.com/cenkalti/backoff instead of rolling our own

operator/pkg/images/images.go
```diff
@@ -0,0 +1,235 @@
...
+	Cassandra_4_0_0:  "datastax/cassandra-mgmtapi-4_0_0:v0.1.12",
+
+	UBICassandra_3_11_6: "datastax/cassandra:3.11.6-ubi7",
+	UBICassandra_3_11_7: "datastax/cassandra:3.11.7-ubi7",
+	UBICassandra_4_0_0:  "datastax/cassandra:4.0-ubi7",
+
+	DSE_6_8_0: "datastax/dse-server:6.8.0",
+	DSE_6_8_1: "datastax/dse-server:6.8.1",
+	DSE_6_8_2: "datastax/dse-server:6.8.2",
+	DSE_6_8_3: "datastax/dse-server:6.8.3",
```
jimdickinson, line 81: 👍  for merging in master already

operator/pkg/images/images.go
```diff
@@ -0,0 +1,235 @@
...
+// 1. Add a new Image enum value below
+// 2. Update imageLookupMap with image url
+// 3. If the image is a Cassandra/DSE image, also update maps:
+//    - versionToOSSCassandra
+//    - versionToDSE
+// 4. Additionally, if there a is a UBI Cassandra/DSE image
+//    available, also update:
+//    - versionToUBIOSSCassandra
+//    - versionToUBIDSE
+//
```
jimdickinson, line 32: 😻

operator/pkg/images/images.go
```diff
@@ -0,0 +1,235 @@
...
+	"6.8.0": UBIDSE_6_8_0,
+	"6.8.1": UBIDSE_6_8_1,
+	"6.8.2": UBIDSE_6_8_2,
+	"6.8.3": UBIDSE_6_8_3,
+}
+
+var log = logf.Log.WithName("images")
+
+func getCustomImageRegistry() string {
+	return os.Getenv(envCustomImageRegistry)
```
jimdickinson, line 124: I'm wondering if the cluster-wide operator request plus the override registry request need to work better together, and if an env var won't work - we should ask @BrynCooke / VMWare

operator/pkg/images/images.go
```diff
@@ -0,0 +1,235 @@
+// Copyright DataStax, Inc.
+// Please see the included license file for details.
+
+package images
```
jimdickinson, line 4: I love splitting this out into its own package, and it's very unit testable 👍 👍 👍 👍 👍

charts/cass-operator-chart/values.yaml
```diff
@@ -9,5 +9,5 @@ webhookClusterRoleName: cass-operator-webhook
 webhookClusterRoleBindingName: cass-operator-webhook
 deploymentName: cass-operator
 deploymentReplicas: 1
-image: "datastax/cass-operator:1.4.0"
+defaultImage: "datastax/cass-operator:1.4.0"
```
jimdickinson, line 12: do we have to change this key?

jimdickinson: also you should get this to 1.4.1 :)

johntrimble: This has actually been in draft and I just got around to merging in master again. It is using the correct image.

It is important for me to be able to distinguish the user specifying an operator image explicitly vs not specifying one. That is reason the I changed `image` -> `defaultImage`. The user will still use `--set image=whatever` to set the operator image.

The general principal here is to respect the user's most specific and explicit instruction for where to get the image. So if the user only specifies the registry, but not the operator image, then we will get the operator image from the registry they gave. If the user gives us both a registry _and_ an image, then we will not modify the image location that they gave us explicitly for the operator. Really, we could just not have the image mentioned at all in the values.yaml, and just have this literal in the deployment.yaml, but I thought it would be easier to find it in this file to update.

operator/pkg/images/images.go
johntrimble: For VMWare, the registry is given to us only when bringing up the operator (it's a setting in the template for the operator). I don't see how an environment variable could not work. If VMWare does a clusterwide install of the operator, then there will only be one such registry for the entire cluster.

charts/cass-operator-chart/values.yaml
johntrimble: I think this block from the deployment.yaml will make it clearer:

```
      - name: cass-operator
        {{- if .Values.image }}
        image: {{ .Values.image }}
        {{- else if .Values.registryName }}
        image: {{ printf "%s/%s" .Values.registryName .Values.defaultImage }}
        {{- else }}
        image: {{ .Values.defaultImage }}
        {{- end }}
```

Maybe there is a better way to do that though?

johntrimble:
Integration tests are not working at present due to issues with k3d/kind. Once that is sorted, I'll run them.

mage/k3d/lib.go
johntrimble: I took this out as we are updating k3d

sandoichi (approved):
great work! :+1:

respringer (approved):
Looks great. +1

jimdickinson (approved):
👍

PR #236, opened 2020-09-08
Better helm charting

respringer:
This adds support for passing an imagePullSecret to the helm chart, so that the operator Docker image may be pulled from that custom Docker registry.

The integration tests have been updated to check for the existence of three environment variables, which if they are found will be used to create an image pull secret.  That secret will then be used by all of the integration tests as the source of the operator Docker image.

This process is described in tests/README.md.


tests/helm_chart_imagepullsecrets/helm_chart_imagepullsecrets_suite_test.go
```diff
@@ -0,0 +1,65 @@
+// Copyright DataStax, Inc.
```
respringer, line 1: This test will probably be removed from the PR soon, because any of the tests can be used to test the custom image pull secret.

README.md
```diff
@@ -204,6 +205,36 @@ kubectl create namespace cass-operator-system
 helm install --set clusterWideInstall=true --namespace=cass-operator-system cass-operator ./charts/cass-operator-chart
 ```
 
+#### Using a custom Docker registry with the Helm Chart
```
jimdickinson, line 208: great additions to the readme!

jimdickinson (approved):
👍  looks great!

charts/cass-operator-chart/templates/deployment.yaml
```diff
@@ -13,6 +13,10 @@ spec:
         name: cass-operator
     spec:
       serviceAccountName: {{ .Values.serviceAccountName }}
+    {{- if .Values.imagePullSecret }}
+      imagePullSecrets:
+      - name: {{ .Values.imagePullSecret }}
+    {{- end }}
```
johntrimble, line 19: So my feeling is that we should also put these image pull secrets in the PodSpec for the statefulsets in case someone wants to use a github package for the config-builder or maybe the management api (though we don't put that on GH packages presently). Note that the statefulsets do not share a service account with the operator, so we would have to have the operator add the pull secrets when creating the statefulsets. Alternatively, we could just make this the user's problem and have them add the image pull secrets to the CassandraDacenter resource themselves (or associate an appropriate service account).

tests/upgrade_operator/upgrade_operator_suite_test.go
```diff
@@ -46,16 +46,13 @@ func TestLifecycle(t *testing.T) {
 func InstallOldOperator() {
 	step := "install old Cass Operator v1.1.0"
 	By(step)
-	err := helm_util.Install(oldOperatorChart, "cass-operator", ns.Namespace, map[string]string{})
-	mageutil.PanicOnError(err)
+	ginkgo_util.HelmInstallWithOverrides(oldOperatorChart, ns.Namespace, map[string]string{})
```
johntrimble, line 49: So calling it in this way will result in the operator image potentially being overriden. This is a case where we want to _ignore_ the M_OPERATOR_IMAGE setting as the purpose of this test is to install an _old_ operator and then upgrade it.

tests/upgrade_operator/upgrade_operator_suite_test.go
```diff
@@ -46,16 +46,13 @@ func TestLifecycle(t *testing.T) {
...
 }
 
 func UpgradeOperator() {
 	step := "upgrade Cass Operator"
 	By(step)
-	var overrides = map[string]string{"image": cfgutil.GetOperatorImage()}
-	err := helm_util.Upgrade("../../charts/cass-operator-chart", "cass-operator", ns.Namespace, overrides)
-	mageutil.PanicOnError(err)
+	ginkgo_util.HelmUpgradeWithOverrides("../../charts/cass-operator-chart", ns.Namespace, map[string]string{})
```
johntrimble, line 55: For the sake of symmetry with `InstallOldOperator`, I would leave this function unchanged. This test is testing the behavior of changing the operator image, so we want to let it be explicit about what it is doing.

johntrimble (commented):
Left some feedback. This is similar to, but not the same as: https://github.com/datastax/cass-operator/pull/228 . I don't think the conflicts will be difficult to reconcile though.

charts/cass-operator-chart/templates/deployment.yaml
jimdickinson: so this is just the cass-operator chart... I don't think we should propagate any secrets from the operator deployment itself into the statefulsets - the user put it here so they can put it in a service account that they put in their CassDC, or we can add `imagePullSecrets` to the CassDC spec

johntrimble: Also, it might be nice to have the user just provide the registry name, username, and password and have the helm chart just make the secret for them. I did something similar for: https://github.com/datastax/cass-operator/pull/228

tests/upgrade_operator/upgrade_operator_suite_test.go
respringer: Ok, I will revert the changes to this test.

PR #238, opened 2020-09-09
Add arm64 support

johntrimble:
This PR:

1. Adds support to build arm64 images of the operator
2. Moves much of the docker build into the workflows
3. Replaces standard `docker build` with `docker buildx build`
4. Updates `mage operator:buildDocker` to use buildkit so that it can still build from the Dockerfile to facilitate testing
5. Merges UBI and standard dockerfiles to better enable buildx
6. Adds separate *-arm64 tags to GH packages
7. Adds release workflow to push multiarch image from ECR to dockerhub


mage/sh/lib.go
```diff
@@ -21,11 +22,16 @@ import (
 // to print output, if not, output will be hidden.
 // Stderr will work as normal
 func Run(cmd string, args ...string) error {
+	return RunWithEnv(nil, cmd, args...)
+}
+
+func RunWithEnv(env map[string]string, cmd string, args ...string) error {
```
johntrimble, line 28: To my knowledge, we had no way to provide environment variables to any of these functions, but I needed this capability to set the `DOCKER_BUILDKIT` environment variable to instruct `docker build` to use buildkit. Not sure if adding a whole new set of functions that take an `env` map, in addition to existing functions, is the right thing to do. Maybe it's better to just refactor everything to always pass the env map.

mage/docker/lib.go
```diff
@@ -165,7 +166,12 @@ func Build(contextDir string, targetStage string, dockerFilePath string, namesAn
 		args = append(args, "--build-arg")
 		args = append(args, x)
 	}
-	return FromArgs(args)
+	c := FromArgs(args)
+	if c.Env == nil {
+		c.Env = map[string]string{}
+	}
+	c.Env["DOCKER_BUILDKIT"] = "1"
```
johntrimble, line 173: The Dockerfile has been updated to use buildx which uses buildkit under the hood. In order to build locally using `docker build` we need to enable buildkit by setting this environment variable. Otherwise, `docker build` will have issues parsing the Dockerfile.

.github/workflows/operatorBuildAndDeploy.yml
```diff
@@ -38,35 +38,143 @@ jobs:
...
+
+
+          # Workaround for GH packages
+
+          docker buildx build \
+            --load \
+            "${STANDARD_ARGS[@]}" \
+            "${GH_TAGS[@]}" \
+            --platform linux/amd64 \
+            .
```
johntrimble, line 156: GH packages don't support multiarch images, or a lot of other things. (This _should_ be fixed in GitHub's new docker registry beta, but I haven't tested it. It does fix containerd though.) This means we have to push images with different tags for different architectures for GH packages. Also, buildx has trouble pushing directly to GH packages... so we have to load the images into docker local images and push them using `docker push`.

.github/workflows/operatorBuildAndDeploy.yml
```diff
@@ -38,35 +38,143 @@ jobs:
...
+            GH_UBI_TAGS+=(--tag "${GH_REPOSITORY}:${t}-ubi")
+            GH_ARM64_TAGS+=(--tag "${GH_REPOSITORY}:${t}-arm64")
+          done
+
+          LABELS=(
+            --label "org.label-schema.schema-version=1.0"
+            --label "org.label-schema.vcs-ref=$GITHUB_SHA"
+            --label "org.label-schema.vcs-url=$GITHUB_REPO_URL"
+            --label "org.label-schema.version=$VERSION"
+          )
```
johntrimble, line 105: This is a quasi-standard for labels. See http://label-schema.org/. We will use the version label here in the release workflow to verify we are creating a release for the correct version.

.github/workflows/release.yml
```diff
@@ -0,0 +1,71 @@
...
+          fi
+
+          # Tag images for DockerHub and push them
+          for t in "$VERSION" "latest"; do
+            docker tag "${ECR_REPOSITORY}:${GITHUB_SHA}" "${DOCKERHUB_REPOSITORY}:${t}"
+            docker push "${DOCKERHUB_REPOSITORY}:${t}"
+
+            docker tag "${ECR_REPOSITORY}:${GITHUB_SHA}-ubi" "${DOCKERHUB_REPOSITORY}:${t}-ubi"
+            docker push "${DOCKERHUB_REPOSITORY}:${t}-ubi"
+          done
```
johntrimble, line 71: I don't have a great way to test this without using a release. I did not feel comfortable putting the ECR secrets into my fork. It is actually pretty straightforward though. We just take the images in ECR and push them to DockerHub. It would be rather difficult to use GitHub packages to do this here due to their lack of multiarch support.

johntrimble:
Depends on https://github.com/datastax/cass-config-builder/pull/19

.github/workflows/operatorBuildAndDeploy.yml
```diff
@@ -38,35 +38,143 @@ jobs:
...
-          MO_GH_USR: 'datastax/cass-operator'
-          MO_GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          MO_GH_PKG_REPO: 'datastax/cass-operator/operator'
-        run: |
+          fi
+
           export PATH=$GOROOT/bin:$GOPATH/bin:$PATH
-          export MO_TAGS=$(cat ./build/tagsToPush.txt)
-          mage operator:deployToGHPackages
+          GITHUB_REPO_URL="https://github.com/${{ github.repository }}"
```
johntrimble, line 76: The following should look much like https://github.com/datastax/cass-config-builder/pull/19/files#diff-a1980bd700940451c303f902e86b140a in cass-config-builder as it does essentially the same thing.

.github/workflows/release.yml
```diff
@@ -0,0 +1,71 @@
...
+          ECR_REPOSITORY="${ECR_REPO}/datastax/cass-operator"
+
+          # Pull the image from ECR so that we can push it to dockerhub.
+          # We exploit the fact that we tag docker images in ECR with the
+          # git sha.
+          docker pull "${ECR_REPOSITORY}:${GITHUB_SHA}"
+          docker pull "${ECR_REPOSITORY}:${GITHUB_SHA}-ubi"
+
+          # Get the version label of the ECR image so that we can double
+          # check that it makes since for this tag name.
```
respringer, line 39: typo "makes sense"

mage/sh/lib.go
respringer: If we end up always using the RunWithEnv versions, we can clean things up in a follow-up ticket that removes the unused functions and possibly removes the WithEnv suffix.

mage/sh/lib.go
```diff
@@ -42,14 +48,22 @@ func RunPanic(cmd string, args ...string) {
...
 }
 
 // Run command and print any output to stdout/stderr
 // Will automatically panic on error
 func RunVPanic(cmd string, args ...string) {
-	err := RunV(cmd, args...)
+	RunVPanicWithEnv(nil, cmd, args...)
+}
+
+func RunVPanicWithEnv(env map[string]string, cmd string, args ...string) {
```
respringer, line 65: There's gotta be some sort of a "panic'd running with envy" joke laying around here somewhere...

respringer (approved):
Looks good, left a couple of comments.  +1

.github/workflows/release.yml
johntrimble: Err, actually, we might need skopeo here... The local images cache does not support multiarch images, so I think if I pull it this way, what I'll end up pushing is just the image for the current platform.

johntrimble: Yep. That's exactly what happens.

johntrimble: Okay, `skopeo copy --all` does what we want.

.github/workflows/release.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Publish to Dockerhub
+        env:
+          ECR_REPO: ${{ secrets.ECR_REPO }}
+        run: |
+          GITHUB_REPO_URL="https://github.com/${{ github.repository }}"
+          GIT_TAG="${GITHUB_REF##*/}"
+          VERSION=${GIT_TAG#v} # strip the initial 'v' from the tag to get the version
+
+          DOCKERHUB_REPOSITORY="datastax/cass-operator"
+          ECR_REPOSITORY="${ECR_REPO}/datastax/cass-operator"
```
johntrimble, line 30: We are going to reference ECR instead of github packages as github packages does not support multiarch images.

.github/workflows/operatorBuildAndDeploy.yml
jimdickinson: should we move long chunks of code here into a proper script in the repo?

operator/docker/base/Dockerfile
```diff
@@ -45,15 +55,65 @@ RUN MO_VERSION=${VERSION_STAMP} mage operator:buildGo
...
 WORKDIR /go
 
 # All we need from the builder image is operator executable
-COPY --from=builder /cass-operator/build/bin/cass-operator-linux-amd64 bin/operator
+ARG TARGETARCH
+COPY --from=builder /cass-operator/build/bin/cass-operator-linux-${TARGETARCH} bin/operator
 
 CMD [ "/go/bin/operator" ]
+
+# UBI Image
```
jimdickinson, line 71: took me a second to figure out what happened, but making one Dockerfile and using the targets is good with me

.github/workflows/release.yml
```diff
@@ -0,0 +1,63 @@
...
+      - 'v*.*.*'
+
+jobs:
+  release_cass_operator:
+    name: Release Cass Operator
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Login Skopeo DockerHub
+        run: echo "${{ secrets.DOCKER_PASSWORD }}" | skopeo login -u "${{ secrets.DOCKER_USERNAME }}" --password-stdin docker.io
```
jimdickinson, line 15: where does `skopeo` come from? it's already part of `ubuntu-latest`?

johntrimble: Skopeo is installed by default, like aws utils. So yes? Not sure what magic gives us those, but it is there, version 1.1.1.

.github/workflows/operatorBuildAndDeploy.yml
johntrimble: That's a good idea. I'll wait until I get the rest of the feedback and then move this into a script (or multiple scripts).

.github/workflows/operatorBuildAndDeploy.yml
```diff
@@ -40,35 +40,143 @@ jobs:
...
+        id: cache
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+      - name: Build and Push Docker
+        if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'datastax/cass-operator'
+        env:
+          ECR_REPO: ${{ secrets.ECR_REPO }}
```
jimdickinson, line 68: looks like this secret is named wrong - ECR_REGISTRY would be better

johntrimble: Technically, it should be registry I think. I cannot change the secret names though or add new ones on this repo or the config builder repo.

jimdickinson: yeah, better to not fuss with it, maybe I'll leave a comment on it in another PR

jimdickinson (approved):
👍  amazing job John!

PR #242, opened 2020-09-09
Support safely scaling down dc/decommissioning nodes

sandoichi:


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -314,14 +314,16 @@ type CassandraStatusMap map[string]CassandraNodeStatus
...
-	DatacenterUpdating       DatacenterConditionType = "Updating"
-	DatacenterStopped        DatacenterConditionType = "Stopped"
-	DatacenterResuming       DatacenterConditionType = "Resuming"
-	DatacenterRollingRestart DatacenterConditionType = "RollingRestart"
+	DatacenterReady           DatacenterConditionType = "Ready"
+	DatacenterInitialized     DatacenterConditionType = "Initialized"
+	DatacenterReplacingNodes  DatacenterConditionType = "ReplacingNodes"
+	DatacenterScalingUp       DatacenterConditionType = "ScalingUp"
+	DatacenterScalingDown     DatacenterConditionType = "ScalingDown"
+	DatacenterScaleDownFailed DatacenterConditionType = "ScaleDownFailed"
```
respringer, line 322: Would we ever need a DatacenterScaleUpFailed ?

operator/deploy/role.yaml
```diff
@@ -11,6 +11,7 @@ rules:
   - services
   - endpoints
   - persistentvolumeclaims
+  - persistentvolumeclaims
```
respringer, line 14: Duplicate "persistentvolumeclaims" looks like a typo.

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -687,17 +688,6 @@ func (rc *ReconciliationContext) CheckRackScale() result.ReconcileResult {
 				return result.Error(err)
 			}
 		}
-
-		currentReplicas := statefulSet.Status.CurrentReplicas
-		if currentReplicas > desiredNodeCount {
```
respringer, line 692: Would it be better to leave this conditional with an additional clause to check if we are not scaling down?

operator/pkg/reconciliation/decommission_node.go
```diff
@@ -0,0 +1,327 @@
...
+					if patchErr != nil {
+						rc.ReqLogger.Error(patchErr, "error patching datacenter status for scale down failed")
+						return patchErr
+					}
+				}
+				return err
+			}
+
+			if err := rc.NodeMgmtClient.CallDecommissionNodeEndpoint(pod); err != nil {
+				// TODO this returns a 500 when it works
```
respringer, line 99: ouch. :)

respringer (approved):
Looks good as-is.  Left a couple of comments. +1

operator/pkg/reconciliation/reconcile_racks.go
sandoichi: So with the current logic, we don't attempt a scale down until the end of the reconcile loop after everything is happy. At this point in the code, we wouldn't have that condition set yet :(

operator/pkg/reconciliation/decommission_node.go
```diff
@@ -0,0 +1,327 @@
...
+			err := fmt.Errorf(msg)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (rc *ReconciliationContext) GetUsedStorageForPods(epData httphelper.CassMetadataEndpoints) (map[string]float64, error) {
+	podStorageMap := make(map[string]float64)
```
johntrimble, line 287: For my own enlightenment, what is the benefit of `make(map[string]float64)` over the literal `map[string]float64{}`?

johntrimble (commented):
It's looking great Chris! I left some feedback.

operator/pkg/reconciliation/decommission_node.go
```diff
@@ -0,0 +1,327 @@
...
+		podRack := pod.Labels[api.RackLabel]
+		if podRack == rackName && strings.HasSuffix(pod.Name, lastPodSuffix) {
+			mgmtApiUp := isMgmtApiRunning(pod)
+			if !mgmtApiUp {
+				return fmt.Errorf("Management API is not up on node that we are trying to decommission")
+			}
+
+			if err := rc.EnsurePodsCanAbsorbDecommData(pod, epData); err != nil {
+				dcPatch := client.MergeFrom(rc.Datacenter.DeepCopy())
+				updated := rc.setCondition(api.NewDatacenterCondition(api.DatacenterScaleDownFailed,
```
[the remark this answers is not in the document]
johntrimble: In the condition message, we probably want to say how much diskspace we need too, as I believe new versions of k8s allow for potentially resizing volumes, so the user might fix the issue that way.

operator/pkg/reconciliation/decommission_node.go
sandoichi: AFAIK there is no difference, other than the `{}` version lets you specify initial elements and the `make()` version lets you specify a capacity. In the case in which you are doing neither, I suppose it is just preference (in which case, I don't mind either way if anyone have a preference).

operator/pkg/reconciliation/decommission_node.go
```diff
@@ -0,0 +1,327 @@
...
+
+		serverDataPv, err := rc.getServerDataPv(pod)
+		if err != nil {
+			return err
+		}
+
+		storageResource := serverDataPv.Spec.Capacity["storage"]
+		total := storageResource.AsDec().UnscaledBig().Int64()
+		used := podsUsedStorage[pod.Name]
+		free := total - int64(used)
```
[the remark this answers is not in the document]
sandoichi: Yeah this is the worst-case scenario. If we want to start out with some fudge instead, then I can change it np.

operator/pkg/reconciliation/decommission_node.go
johntrimble: Ah okay, I have no preference either, I was just curious.

operator/pkg/reconciliation/decommission_node.go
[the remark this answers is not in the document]
sandoichi: Ah, that sounds like a nice way to do things. Thanks for the detailed write up!

mage/ginkgo/lib.go
```diff
@@ -466,3 +467,37 @@ func (ns NsWrapper) RetrieveStatusFromNodetool(podName string) []NodetoolNodeInf
 	}
 	return nodeInfo
 }
+
+func (ns NsWrapper) RetrieveSuperuserCreds(clusterName string) (string, string) {
+	secretName := fmt.Sprintf("%s-superuser", clusterName)
+	secretResource := fmt.Sprintf("secret/%s", secretName)
```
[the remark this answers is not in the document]
sandoichi: Doh, didn't think about custom creds in integ tests. I'll make it support that.

operator/pkg/reconciliation/decommission_node.go
```diff
@@ -0,0 +1,327 @@
...
+
+// Wait for decommissioning nodes to finish before continuing to reconcile
+func (rc *ReconciliationContext) CheckDecommissioningNodes(epData httphelper.CassMetadataEndpoints) result.ReconcileResult {
+	if rc.Datacenter.GetConditionStatus(api.DatacenterScalingDown) == corev1.ConditionFalse {
+		return result.Continue()
+	}
+
+	for _, pod := range rc.dcPods {
+		if pod.Labels[api.CassNodeState] == stateDecommissioning {
+			if !IsDoneDecommissioning(pod, epData) {
```
johntrimble, line 130: Don't forget to clear the pod from `CassandraDatacenter.Status.NodeStatuses`

sandoichi: Thanks for the reminder :)

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -1361,9 +1351,7 @@ func (rc *ReconciliationContext) CheckDcPodDisruptionBudget() result.ReconcileRe
 	return result.Continue()
 }
 
-// UpdateRackNodeCount ...
```
jimdickinson, line 1364: don't delete this - make a better docstring - the standards/idioms/linter really wants exported stuff to have a docstring

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -2081,6 +2095,10 @@ func (rc *ReconciliationContext) ReconcileAllRacks() (reconcile.Result, error) {
 		return recResult.Output()
 	}
 
+	if recResult := rc.CheckDecommissioningNodes(endpointData); recResult.Completed() {
```
jimdickinson, line 2098: how did you pick this spot in the order for this part?

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -2137,6 +2155,10 @@ func (rc *ReconciliationContext) ReconcileAllRacks() (reconcile.Result, error) {
 		return recResult.Output()
 	}
 
+	if recResult := rc.DecommissionNodes(endpointData); recResult.Completed() {
```
jimdickinson, line 2158: why not move this up before CheckRackPodTemplate?

operator/pkg/reconciliation/reconcile_racks.go
sandoichi: I needed it to happen after the rack labels got set up

operator/pkg/reconciliation/reconcile_racks.go
sandoichi: initially I put it after everything based on the idea of "we should make sure that everything else is in good shape before we attempt to decommission" but if that idea is fulfilled before we get to `CheckRackPodTemplate` then I'll move it up.

operator/pkg/reconciliation/decommission_node.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+	}
+
+	// this shouldn't happen
+	return fmt.Errorf("Could not find pod to decommission on rack %s", rackName)
+}
+
+// Wait for decommissioning nodes to finish before continuing to reconcile
+func (rc *ReconciliationContext) CheckDecommissioningNodes(epData httphelper.CassMetadataEndpoints) result.ReconcileResult {
+	if rc.Datacenter.GetConditionStatus(api.DatacenterScalingDown) == corev1.ConditionFalse {
```
johntrimble, line 113: What if the condition is Unknown?

sandoichi: Naturally I checked all of the cases in reconcile rack but not in my own file

operator/pkg/reconciliation/reconcile_racks.go
johntrimble: If we put this anywhere after the `CheckCassandraNodeStatuses()`, we'll know that all the pods are up, running, and ready. That works for knowing the pods in the cluster are in a happy state. Where we put this call after that call tells you what operations might get priority over decomissioning, and those that may interrupt decomissioning. For example, if you put the call after `CheckRollingRestart()`, then in-between decomissioning two nodes, a rolling restart might happen.

johntrimble: By the way, the call is probably here because of advice I gave earlier, but your right, it probably should be moved. I don't much care for how other datacenter-wide operations can get interleaved with each other but putting the call here doesn't actually solve that. Also, maybe interleaving is fine.

mage/ginkgo/lib.go
```diff
@@ -466,3 +467,37 @@ func (ns NsWrapper) RetrieveStatusFromNodetool(podName string) []NodetoolNodeInf
...
+func (ns NsWrapper) CqlExecute(podName string, stepDesc string, cql string, user string, pw string) {
+	k := kubectl.ExecOnPod(
+		podName, "--", "cqlsh",
+		"--user", user,
+		"--password", pw,
+		"-e", cql).
+		WithFlag("container", "cassandra")
+	ginkgo.By(stepDesc)
+	ns.ExecVPanic(k)
+}
```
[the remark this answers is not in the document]
jimdickinson: ^

operator/pkg/reconciliation/decommission_node.go
jimdickinson: did you get to test against a fixed version of the management API?

operator/pkg/reconciliation/decommission_node.go
```diff
@@ -0,0 +1,355 @@
+package reconciliation
```
jimdickinson, line 1: 💯 for making a new file for this

jimdickinson (commented):
The test code in this is all really excellent 💯

mage/ginkgo/lib.go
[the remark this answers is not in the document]
johntrimble: We can always add it later.

operator/pkg/reconciliation/decommission_node.go
johntrimble: I'd just make a ticket for us to come back and clean this up later. Make it part of the psp epic.

johntrimble (approved):
Great work on this @sandoichi ! This looks ready to merge to me.

Issue #255, opened 2020-09-16
Support single-node upgrade


grighetto:
As part of a C* QA effort, we'd like to upgrade a single node in a C* cluster in order to test mixed-version scenarios.
We're using the cass-operator to spin up the test clusters, but to my knowledge the operator only supports single rack upgrades, that is, the entire rack.

jsanda:
>to my knowledge the operator only supports single rack upgrades, that is, the entire rack.

I believe this is correct. cass-operator creates a StatefulSet per rack and uses the default update strategy which is `RollingUpdate`.  Let's say you change the `.spec.serverVersion` of your CassandraDatacenter. This will roll out the new image across all StatefulSets/racks in the cluster. You can limit the update to a single rack with `.spec.canaryUpgrade: true`.

The rolling update strategy of the StatefulSet has a `partition` property that allows you to control how many pods get updated. Only those pods whose ordinal index is >= `partition` will get updated. 

I like how CassKop handles updates. Like cass-operator it has a `Rack` type which exposes the partition property. By exposing the partition property at the rack level not only would you have fine-grained control over how many pods are updated, but you also have control over which racks get updated.

We would need to figure out what the behavior should be if both `.spec.canaryUpgrade` is `true` and the partition property at the rack level is modified. That should be easy enough though. 

I will work on a PR for this.

jsanda:
After thinking about it some more, I think the `.spec.canaryUpgrade` property should be removed. By adding the `Partition` property as I have in PR #256 there is no need for `.spec.canaryUpgrade`. Removing it also simplifies the code a bit which is an added bonus.

jimdickinson:
I disagree with removing `canaryUpgrade` for a few reasons. I think what you've added should be used in concert with it. Will leave more comments in the PR.

jsanda:
Hi @grighetto 

I talked some @jimdickinson and he explained to me the idea of adding a `CanaryUpgradeCount` property. Since it would not be defined at the rack level, you would not be able to control in which rack the upgrade happens as I described above. You would however be able to upgrade a single node though. 

Is this sufficient for your use case?

grighetto:
> Since it would not be defined at the rack level, you would not be able to control in which rack the upgrade happens

@jsanda yes, that totally works for me as long as the order of execution is consistent, that is, the upgrade doesn't happen at random racks at different times, such that the tests remain reproducible.

Thanks for the quick turnaround on this.

jsanda:
[...]

PR #256, opened 2020-09-17
add Partition property to Rack to improve canary ugprade support

jsanda:
This is for #255.

### DIFF
--- a/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -214,6 +214,8 @@ type Rack struct {
 	Name string `json:"name"`
 	// Zone name to pin the rack, using node affinity
 	Zone string `json:"zone,omitempty"`
+
+	Partition int32 `json:"partition,omitempty"`
 }
 
 type CassandraNodeStatus struct {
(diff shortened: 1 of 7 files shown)


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -214,6 +214,8 @@ type Rack struct {
 	Name string `json:"name"`
 	// Zone name to pin the rack, using node affinity
 	Zone string `json:"zone,omitempty"`
+
+	Partition int32 `json:"partition,omitempty"`
```
jimdickinson, line 218: I don't think we should put this on the rack. It's just a footgun waiting to happen and there's nothing much to be gained by letting it vary per rack. Let's add CanaryUpgradeCount for the number of pods to roll forward, right next to CanaryUpgrade. If it's missing - the default - you get the current behavior for CanaryUpgrade.

jsanda: >I don't think we should put this on the rack. It's just a footgun waiting to happen

Can you expand on why it is a bad idea to put it on the rack?

>there's nothing much to be gained by letting it vary per rack.

Letting vary by rack allows you to upgrade a single node and it also gives you the flexibility to choose the rack. Do we get that with `CanaryUpgradeCount`?

PR #257, opened 2020-09-18
DSE 6.8.4 support

msmygit:
DSE 6.8.4 support


tests/testdata/default-three-rack-four-node-dc.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   clusterName: cluster1
   serverType: dse
-  serverVersion: "6.8.2"
+  serverVersion: "6.8.4"
```
jimdickinson, line 8: great find!

tests/testdata/default-three-rack-four-node-limited-storage-dc.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   clusterName: cluster1
   serverType: dse
-  serverVersion: "6.8.2"
+  serverVersion: "6.8.4"
```
jimdickinson, line 8: here too

jimdickinson (approved):
looks good! sorry for the delays checking out the int test infra

respringer (approved):

PR #259, opened 2020-09-20
Filter node changes to only taint changes

johntrimble:
To reduce calls to reconcile, only trigger reconcile for node changes that
result from a change of the taints on that node.

### DIFF
--- a/operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
+++ b/operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
@@ -164,7 +164,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 				return true
 			}
 
-			return !utils.DeepEqualArrayIgnoreOrder(
+			return !utils.ElementsMatch(
 				nodeOld.Spec.Taints, nodeNew.Spec.Taints)
 		},
 	}
--- a/operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
+++ b/operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
@@ -146,12 +146,36 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 			return requests
 		})
 
+	nodeTaintsChangedPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			nodeOld, ok1 := e.ObjectOld.(*corev1.Node)
+			nodeNew, ok2 := e.ObjectNew.(*corev1.Node)
+
+			if !ok1 || !ok2 {
+				log.Error(nil, "Failed to cast update.Event objects to type Node", "objectOld", e.ObjectOld, "objectNew", e.ObjectNew)
+				return true
+			}
+
+			if nodeOld == nil && nodeNew == nil {
+				return false
+			}
+
+			if (nodeOld == nil && nodeNew != nil) || (nodeOld != nil && nodeNew == nil) {
+				return true
+			}
+
+			return !utils.DeepEqualArrayIgnoreOrder(
+				nodeOld.Spec.Taints, nodeNew.Spec.Taints)
+		},
+	}
+
 	if utils.IsPSPEnabled() {
 		err = c.Watch(
 			&source.Kind{Type: &corev1.Node{}},
 			&handler.EnqueueRequestsFromMapFunc{
 				ToRequests: nodeMapFn,
 			},
+			nodeTaintsChangedPredicate,
 		)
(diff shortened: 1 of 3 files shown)


respringer (approved):
Looks good. +1

operator/pkg/utils/utilities.go
```diff
@@ -6,13 +6,64 @@ package utils
...
 }
 
+func RangeInt(min, max, step int) []int {
+	size := int(math.Ceil(float64((max - min)) / float64(step)))
+	l := make([]int, size)
+	for i := 0; i < size; i++ {
+		l[i] = min + i * step
+	}
+	return l
+}
```
jimdickinson, line 25: this is one where there's been folks debating it in the golang community many times, and the idiom is just use a for loop inlined...
```
for i := min; i < max; i += step {
    doSomething(i)
}
```

operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
```diff
@@ -146,12 +146,36 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 			return requests
 		})
 
+	nodeTaintsChangedPredicate := predicate.Funcs{
```
jimdickinson, line 149: very much a nit - this is going to be our predicate for all Node UpdateEvents, right? so I might change this name

operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
```diff
@@ -146,12 +146,36 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			nodeOld, ok1 := e.ObjectOld.(*corev1.Node)
+			nodeNew, ok2 := e.ObjectNew.(*corev1.Node)
+
+			if !ok1 || !ok2 {
+				log.Error(nil, "Failed to cast update.Event objects to type Node", "objectOld", e.ObjectOld, "objectNew", e.ObjectNew)
+				return true
```
jimdickinson, line 156: shouldn't this be false?

operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
```diff
@@ -146,12 +146,36 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			}
+
+			if nodeOld == nil && nodeNew == nil {
+				return false
```
jimdickinson, line 160: I think if ok1 and ok2 are both true, these nodes must not be nil?

operator/pkg/utils/utilities.go
```diff
@@ -6,13 +6,64 @@ package utils
...
+
+func isArrayOrSlice(a interface{}) bool {
+	t := reflect.TypeOf(a)
+	k := t.Kind()
+	return k == reflect.Slice || k == reflect.Array
+}
+
+func DeepEqualArrayIgnoreOrder(a interface{}, b interface{}) bool {
```
jimdickinson, line 33: how do you feel about keeping the naming matching testify?

https://github.com/emou/testify/blob/8d9174bb2b778c662039a4dc921b505d14307a6b/assert/assertions.go#L738

operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
johntrimble: If for some reason I can't cast these, I just return true to go ahead and reconcile. It should not happen.

operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
johntrimble: the interface value might be (nil, corev1.Node). In which case it can be cast to corev1.Node and still be nil with ok1 and ok2 being true.

operator/pkg/utils/utilities.go
johntrimble: Sounds good.

operator/pkg/utils/utilities.go
johntrimble: I actually need a slice of ints. See the code where I use this.

operator/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
johntrimble: I named the predicate what it is checking for. Much like the GenerationChangedPredicate and managedByCassandraOperatorPredicate predicates.

jimdickinson (approved):
👍

PR #263, opened 2020-09-25
Podtemplatespec

respringer:
Allow the user to override values in the init-container and three containers generated by cass-operator.  There are more details in the README.md.


operator/pkg/reconciliation/construct_service.go
```diff
@@ -0,0 +1,197 @@
+// Copyright DataStax, Inc.
```
respringer, line 1: There should be no changes to the logic in this file.  I simply moved service related functions to a new file.

operator/pkg/reconciliation/construct_statefulset.go
```diff
@@ -0,0 +1,162 @@
+// Copyright DataStax, Inc.
```
respringer, line 1: Just like construct_service.go, there should be no logic changes in this file either.

operator/pkg/reconciliation/constructor.go
```diff
@@ -6,351 +6,18 @@ package reconciliation
 // This file defines constructors for k8s objects
```
respringer, line 6: This file should now contain miscellaneous construction logic for everything that does not have a dedicated construct_foo.go file.

README.md
```diff
@@ -259,6 +259,57 @@ The operator is comprised of the following container images working in concert:
   4.0-beta1.
 * ... or DSE, built from [datastax/docker-images](https://github.com/datastax/docker-images).
 
+### Overriding properties of cass-operator created Containers
```
jimdickinson, line 262: excellent README addition!

operator/pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -0,0 +1,481 @@
...
+					},
+				},
+			},
+		},
+	}
+}
+
+// calculatePodAntiAffinity provides a way to keep the db pods of a statefulset away from other db pods
+func calculatePodAntiAffinity(allowMultipleNodesPerWorker bool) *corev1.PodAntiAffinity {
+	if allowMultipleNodesPerWorker {
```
jimdickinson, line 53: maybe we should hoist this arg and the if out of here and to the call site? would probably be more clear to read, right?

operator/pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -0,0 +1,481 @@
...
+}
+
+func combineVolumeMountSlices(defaults []corev1.VolumeMount, overrides []corev1.VolumeMount) []corev1.VolumeMount {
+	out := append([]corev1.VolumeMount{}, overrides...)
+outerLoop:
+	// Only add the defaults that don't have an override
+	for _, volumeDefault := range defaults {
+		for _, volumeOverride := range overrides {
+			if volumeDefault.Name == volumeOverride.Name {
+				continue outerLoop
```
jimdickinson, line 124: let's use a `found := false` boolean instead of the named continue

respringer: Ok.  I had that initially and it is more boilerplate-ish, but I can easily switch back.

operator/pkg/reconciliation/construct_podtemplatespec.go
respringer: This is pre-existing code that I didn't change.  Can we update it in a new PR?

operator/pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -0,0 +1,481 @@
...
+}
+
+func combineEnvSlices(defaults []corev1.EnvVar, overrides []corev1.EnvVar) []corev1.EnvVar {
+	out := append([]corev1.EnvVar{}, overrides...)
+outerLoop:
+	// Only add the defaults that don't have an override
+	for _, envDefault := range defaults {
+		for _, envOverride := range overrides {
+			if envDefault.Name == envOverride.Name {
+				continue outerLoop
```
jimdickinson, line 169: same here and the other two places

jimdickinson: nvm

operator/pkg/reconciliation/construct_podtemplatespec.go
jimdickinson: nvm

operator/pkg/reconciliation/reconcile_reaper.go
```diff
@@ -15,52 +15,60 @@ import (
...
+	ReaperContainerName     = "reaper"
+	ReaperHealthCheckPath   = "/healthcheck"
+	ReaperKeyspace          = "reaper_db"
+	ReaperSchemaInitJob     = "ReaperSchemaInitJob"
 	// This code currently lives at https://github.com/jsanda/create_keyspace.
 	ReaperSchemaInitJobImage = "jsanda/reaper-init-keyspace:latest"
 )
 
-func buildReaperContainer(dc *api.CassandraDatacenter) corev1.Container {
+func buildReaperContainer(dc *api.CassandraDatacenter, container *corev1.Container) {
```
jimdickinson, line 35: let's add a docstring to explain why we're modifying the input instead of returning something

jimdickinson: also let's add a note about adding a unit test here - but I know @jsanda has plans to change this code so no need to do it now

jimdickinson (approved):
just a couple small points - looks good to me!

bradfordcp:
Legit awesome @respringer

PR #264, opened 2020-09-28
Fixed a small logic error with conditions

respringer:
Also increased test timeouts to compensate for how slow the reaper test runs in my local dev environment.  The reaper test now passes for me.


jimdickinson (approved):
👍  great find @respringer and @sandoichi !

sandoichi (approved):
:100:

PR #269, opened 2020-10-02
Termination grace time and prestop drain

respringer:
Now the default TerminationGracePeriodSeconds is set to 2 minutes.  Also, by default, the "cassandra" container will have lifecycle prestop hook that calls node drain.  This call can handle both "insecure" and "MTLS" management API modes.


jimdickinson (changes requested):
should `git rm operator/pkg/utils/test-jks`

tests/timeout_prestop_termination/timeout_prestop_termination_suite_test.go
```diff
@@ -0,0 +1,122 @@
...
+			step = "checking that the dc no longer exists"
+			json = "jsonpath={.items}"
+			k = kubectl.Get("CassandraDatacenter").
+				WithLabel(dcLabel).
+				FormatOutput(json)
+			ns.WaitForOutputAndLog(step, k, "[]", 300)
+
+			elapsedTime := startTime.Sub(time.Now())
+
+			Expect(elapsedTime.Minutes() <= 3).To(BeTrue(), "Stop should have completed in under 3 minutes")
```
jimdickinson, line 98: 👍

respringer:
> should `git rm operator/pkg/utils/test-jks`

Arg, why does that file even get created?  I am usually pretty good at explicitly ignoring it, but we should either prevent it from being made or add it to the gitignore.  I will take a look.

jimdickinson:
> > should `git rm operator/pkg/utils/test-jks`
> 
> Arg, why does that file even get created? I am usually pretty good at explicitly ignoring it, but we should either prevent it from being made or add it to the gitignore. I will take a look.

No worries! Yes feel free to add it to .gitignore - I hadn't seen this particular file in this location before

jimdickinson (approved):
looks very good!

tests/testdata/default-single-rack-single-node-prestop-dc.yaml
```diff
@@ -0,0 +1,36 @@
...
+    insecure: {}
+  size: 1
+  podTemplateSpec:
+    spec:
+      containers:
+        - name: "cassandra"
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sleep", "6000s"]
```
respringer, line 19: This is to force the shutdown to take forever so the grace period timeout will kick-in.

operator/pkg/utils/crypto_test.go
```diff
@@ -79,6 +80,10 @@ func Test_GetJKS(t *testing.T) {
 	if len(jks) == 0 {
 		t.Errorf("JKS blob too small")
 	}
-	ioutil.WriteFile("test-jks", jks, 0644)
 
+	defer func() {
+		_ = os.Remove("test-jks")
```
sandoichi, line 85: :+1:

sandoichi (approved):
:100:

PR #272, opened 2020-10-06
fix the loading of dev images during integ test setup

sandoichi:


buildsettings.yaml
```diff
@@ -23,6 +23,6 @@ dev:
   ubiDseImages:
     - datastax/dse-server:6.8.4-ubi7
   ubiOssImages:
-    - datastax/cassandra:3.11.7-ubi7
+    - datastax/cassandra:3.11.6-ubi7
```
sandoichi, line 26: 3.11.7-ubi7 is not in dockerhub, so this was breaking integ tests when loading images

jimdickinson: cc @bradfordcp

jimdickinson (approved):
👍

respringer (approved):

Issue #66, opened 2020-04-29
Run cassandra pods as nonRootUser


joe-sch:
It should be possible to define a different user than root in a  cassandradatacenter.

When i exec into a cassandra pod, i see that the pod is executed as root user.
But many k8s clusters have a policy which do not allow to run an image as root user.
In the crd definition i can't find an option to define a securityContext for the pods.

jimdickinson:
Do you have a cassandra docker image you've used previously that doesn't come up as root? I spent quite a bit of time looking at this because it's one of the few differences in the API for us to bring up DSE pods vs. Cassandra pods. The upstream images from https://github.com/docker-library/cassandra do start as root, and then they drop to a cassanda user using gosu just before Java starts. ([link](https://github.com/docker-library/cassandra/blob/f6829f33cac880b0f0787b223d3a8b19f82efab5/docker-entrypoint.sh#L14))

We intend for users to use cassandra images from https://github.com/datastax/management-api-for-apache-cassandra and we based our image on , but we do want to support anyone forking that and making minimal changes, like removing the possibility of using the root user. Let us know more about your configuration / requirements and we'll make an update to unblock you!

cc @eldondevat @tjake

joe-sch:
Yes we have cassandra running in a dev environment (production is planned for autumn), 
Our statefulset is based on this example: https://kubernetes.io/docs/tutorials/stateful-application/cassandra/. But we added a `securityContext` to run the cassandra pods as user 999.
The user and group with id 999 is built in: https://github.com/docker-library/cassandra/blob/master/3.11/Dockerfile 

````
securityContext:
        runAsUser: 999
````

This is ok for us and help us to run on our on prem kubernetes envirenment.
However, it would be ideal if the user could choose freely, as described by [this guide](https://docs.openshift.com/container-platform/4.1/openshift_images/create-images.html) `SUPPORT ARBITRARY USER IDS`

PR #275, opened 2020-10-08
Setup securitycontext for cassandra

respringer:
This sets up the SecurityContext for Cassandra in the same way that it was set up for DSE.  This is related to: 

https://github.com/datastax/management-api-for-apache-cassandra/pull/43

### DIFF
--- a/operator/pkg/reconciliation/construct_statefulset.go
+++ b/operator/pkg/reconciliation/construct_statefulset.go
@@ -125,13 +125,11 @@ func newStatefulSetForCassandraDatacenterHelper(
 	}
 
 	// workaround for https://cloud.google.com/kubernetes-engine/docs/security-bulletins#may-31-2019
-	if dc.Spec.ServerType == "dse" {
-		var userID int64 = 999
-		template.Spec.SecurityContext = &corev1.PodSecurityContext{
-			RunAsUser:  &userID,
-			RunAsGroup: &userID,
-			FSGroup:    &userID,
-		}
+	var userID int64 = 999
+	template.Spec.SecurityContext = &corev1.PodSecurityContext{
+		RunAsUser:  &userID,
+		RunAsGroup: &userID,
+		FSGroup:    &userID,
 	}
 
 	_ = httphelper.AddManagementApiServerSecurity(dc, template)


jimdickinson:
fixes #66

jimdickinson:
We might need to make sure we get your fix into a 3.11.6 image, too. But this looks great!

jimdickinson (approved):

operator/pkg/reconciliation/construct_statefulset.go
```diff
@@ -125,7 +126,15 @@ func newStatefulSetForCassandraDatacenterHelper(
...
-	if dc.Spec.ServerType == "dse" {
+
+	serverImage := ""
+	for _, c := range template.Spec.Containers {
+		if c.Name == CassandraContainerName {
+			serverImage = c.Image
+		}
+	}
+
+	if dc.Spec.ServerType == "dse" || images.DockerImageRunsAsCassandra(dc.Spec.ServerVersion, serverImage) {
```
johntrimble, line 137: What if a user explicitly specifies an image? How do we know if it uses the cassandra user or not? What if its image name doesn't follow the same regex?

respringer: Good catch.  I just added a commit with a test case with an unparseable image name that defaults to assuming cassandra is running as root.

johntrimble: My feeling is that we should default to the thing we want to be true (which I assume is use the Cassandra user). Like what if I'm testing some version of the cassandra image still in dev (where my tag might just be a git hash)? Do we need a flag to permit the user to make explicit whether or not the cassandra user is in use to override inferring from the image string?

respringer: I suggested a flag like that at one point.

How about:

1. If we don't recognize the image name, we default to assuming the image runs as cassandra
2. We look for CASSOP_RUN_AS_CASSANDRA="false" to override the default?

@jimdickinson ^ ?

respringer: Ok, after discussing this in standup, I have provided a DockerImageRunsAsCassandra boolean in the Datacenter spec.  It is optional and can be left undefined.  If it is defined and set to true, then we will define the security context.  

Otherwise we fall back to the calculation logic in images.CalculateDockerImageRunsAsCassandra()

Note that CalculateDockerImageRunsAsCassandra() will default to true if it cannot recognize the image and version combination, which will cause the securitycontext to be defined.

PR #279, opened 2020-10-09
Added kubebuilder:pruning:PreserveUnknownFields

respringer:


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -79,6 +79,7 @@ type CassandraDatacenterSpec struct {
 	ServerType string `json:"serverType"`
 
 	// Config for the server, in YAML format
+	// +kubebuilder:pruning:PreserveUnknownFields
```
sandoichi, line 82: :100:

sandoichi (approved):

charts/cass-operator-chart/templates/customresourcedefinition.yaml
```diff
@@ -6307,9 +6307,7 @@ spec:
               type: string
           type: object
       type: object
-      {{- if semverCompare ">= 1.15-0" .Capabilities.KubeVersion.GitVersion }}
```
jimdickinson, line 6310: I think we need to keep this if we want to support k8s <= 1.14 - which we might be OK moving away from

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
jimdickinson: 🎉

charts/cass-operator-chart/templates/customresourcedefinition.yaml
respringer: @jimdickinson should I put the conditional back?

jimdickinson: for this PR, I think so... I thought we had some more hackery around getting `x-kubernetes-preserve-unknown-fields: true` into the yaml, but maybe that had already been cleaned up and turned into "apply the patch"

PR #285, opened 2020-10-15
Win OS watch of kubectl get pods (K8C-9)

jeffbanks:
Task reference: [K8C-9](https://k8ssandra.atlassian.net/browse/K8C-9)

Noticed during local testing with k8s:setupCassandraCluster that the Linux `watch `command was being used, which isn't supported directly on Win OS.  

A 3-second timeout included (with ability to advance by user if wanting it more frequently) before issuing `get pods` again.

```
Waiting for 0 seconds, press a key to continue ...
NAME                             READY   STATUS    RESTARTS   AGE
cass-operator-65bc9494b8-xtvjk   1/1     Running   0          8m59s
cluster1-dc1-default-sts-0       2/2     Running   0          8m39s
cluster1-dc1-default-sts-1       2/2     Running   0          8m39s
cluster1-dc1-default-sts-2       2/2     Running   0          8m39s
```

Test class created for additional coverage to be added.  Need to determine best approach for decoupling the two utility dependencies (mage and sh) from the `kubectl` module.  One thought is to use a similar approach that is used for `KCmd` struct.


tests/kube_ctl/watch_pods_test.go
```diff
@@ -0,0 +1,29 @@
+package kube_ctl
```
jimdickinson, line 1: what was this test about?

jeffbanks: At the time, I was thinking of coverage for kubectl `WatchPods()` and `WatchPodsInNs()` to ensure OS specific commands are issued to `RunVPanic`(...).  If you still think this would be good to have, I can add it to my list.  If not, we can remove this test setup.

jeffbanks:
> I'm not sure the context, but can't we just use `-w` / `--watch` option of `kubectl get`? It works on my windows.

Good suggestion, I wasn't familiar with kubectl at the time of this PR.  Let me give that a try locally to verify.  That would make this a less OS specific solution.

jeffbanks:
Removed the shell of this test as we are able to use the `kubectl` with `-w` option instead of making an OS specific check.  

Thanks for the suggestion here @yukim 

The original issue appears to be sending out a command that was not based on `kubectl`, but rather a Linux `watch` command, which is not supported by Win OS directly.

jimdickinson (approved):

PR #288, opened 2020-10-19
KO-523 permit pods on tainted node

johntrimble:


operator/pkg/psp/emm.go
```diff
@@ -0,0 +1,688 @@
...
+	NotEnoughResources      EMMFailure = "NotEnoughResources"
+	TooManyExistingFailures EMMFailure = "TooManyExistingFailures"
+)
+
+// This interface is what, in practice, the ReconciliationContext will need to
+// implement. Its intention is to keep coupling with the reconciliation package
+// low and also to keep this package relatively sheltered from certain thorny
+// details, like determining which pods represent cassandra nodes that have 
+// bootstrapped at some point.
+type EMMSPI interface {
```
johntrimble, line 37: Actually, it just makes testing easier when you have an interface.

operator/pkg/psp/emm.go
```diff
@@ -0,0 +1,688 @@
...
+	removeNextPodFromEvacuateDataNode() (bool, error)
+	removeAllPodsFromPlannedDowntimeNode() (bool, error)
+	startNodeReplace(podName string) error
+}
+
+type EMMService interface {
+	EMMOperations
+	EMMChecks
+	getLogger() logr.Logger
+}
```
johntrimble, line 78: These interfaces make testing easier.

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -2181,15 +2366,6 @@ func (rc *ReconciliationContext) ReconcileAllRacks() (reconcile.Result, error) {
...
 	}
 
-	// We do the node and pvc taint checks here
-	// with the assumption that the cluster is healthy
-
-	if utils.IsPSPEnabled() {
-		if err := rc.checkNodeAndPvcTaints(); err != nil {
-			return result.Error(err).Output()
-		}
-	}
```
johntrimble, line 2191: Since we are now permitting not ready pods on nodes undergoing EMM, this check has been moved.

tests/psp_emm/psp_emm_suite_test.go
```diff
@@ -45,54 +62,30 @@ func TestLifecycle(t *testing.T) {
...
 			}()
 
-			// node.vmware.com/drain=planned-downtime:NoSchedule
-			step = fmt.Sprintf("tainting node: %s", node1Name)
+			step := fmt.Sprintf("tainting node: %s", node1Name)
 			k = kubectl.Taint(
 				node1Name,
 				"node.vmware.com/drain",
-				"planned-downtime",
+				"drain",
```
johntrimble, line 88: Switched this to drain since it better matches the what this was testing (deleting the pvc).

tests/psp_emm/psp_emm_suite_test.go
```diff
@@ -45,54 +62,30 @@ func TestLifecycle(t *testing.T) {
...
-				k = kubectl.PatchMerge(node1Resource, json)
-				err = k.ExecV()
-				if err != nil {
-					panic(err)
-				}
+				k := kubectl.Taint(
+					node1Name,
+					"node.vmware.com/drain",
+					"",
+					"NoSchedule-")
```
johntrimble, line 80: I think this is more straightforward.

tests/psp_pvc_annotation/psp_pvc_annotation_suite_test.go
```diff
@@ -82,22 +82,26 @@ var _ = Describe(testName, func() {
...
-			for i < 300 {
-				time.Sleep(1 * time.Second)
-				i += 1
-
-				names := ns.GetDatacenterReadyPodNames(dc1Name)
-				if len(names) < 3 {
-					break
-				}
-			}
+			ns.WaitForDatacenterReadyPodCount(dc1Name, 2)
```
johntrimble, line 85: I think this check does what the previous code was trying to accomplish.

operator/pkg/psp/emm.go
```diff
@@ -0,0 +1,681 @@
...
+	nodes2, err := impl.getNodeNameSetForPlannedDownTime()
+	if err != nil {
+		return false, err
+	}
+	nodes = utils.UnionStringSet(nodes, nodes2)
+
+	// Strip EMM failure annotation from pods where node is no longer tainted
+	podsFailedEmm := impl.getPodsWithAnnotationKey(EMMFailureAnnotation)
+	nodesWithPodsFailedEMM := utils.GetNodeNameSetForPods(podsFailedEmm)
+	nodesNoLongerEMM := utils.SubtractStringSet(
```
respringer, line 142: If we ever want to polish things further, we could abstract a stringset library.

respringer (approved):
There's a lot of logic here, but it looks good and the unit tests look comprehensive.  +! as-is.  Made a couple of minor rename suggestions.

operator/pkg/psp/emm_test.go
```diff
@@ -0,0 +1,322 @@
...
+type MockEMMService struct {
+	mock.Mock
+}
+
+func (m *MockEMMService) getRacksWithNotReadyPodsBootstrapped() []string {
+	args := m.Called()
+	return args.Get(0).([]string)
+}
+
+func (m *MockEMMService) getNodeNameSetForRack(rackName string) map[string]bool {
```
jimdickinson, line 43: many uses of `map[string]bool` - can we do something like
```
type map[string]bool StringSet
```
for more readable code?

jimdickinson: (just left a comment on the first one that comes up on GH, nothing special about this one)

operator/pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -474,6 +474,18 @@ func buildPodTemplateSpec(dc *api.CassandraDatacenter, zone string, rackName str
 	}
 	baseTemplate.Labels = utils.MergeMap(baseTemplate.Labels, podLabels)
 
+	// Annotations
+
+	podAnnotations := map[string]string{}
+	if utils.IsPSPEnabled() {
+		podAnnotations["vsphere-emm-integrated"] = "true"
```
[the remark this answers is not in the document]
johntrimble: You didn't miss anything, this requirement was added after your work.

operator/pkg/psp/emm.go
```diff
@@ -0,0 +1,681 @@
...
+
+	evacuateDataNameSet, err := impl.getNodeNameSetForEvacuateAllData()
+	if err != nil {
+		return false, err
+	}
+
+	taintedNodesNameSet := utils.UnionStringSet(plannedDownNameSet, evacuateDataNameSet)
+	podsNotReadyOnTaintedNodes := utils.FilterPodsWithNodeInNameSet(podsNotReady, taintedNodesNameSet)
+	
+	if len(podsNotReadyOnTaintedNodes) > 1 {
```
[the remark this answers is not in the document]
johntrimble: Hmm... how is this even working at all. I will take a look, thanks for catching this.

operator/pkg/reconciliation/reconcile_racks.go
```diff
@@ -2135,6 +2296,16 @@ func (rc *ReconciliationContext) ReconcileAllRacks() (reconcile.Result, error) {
...
 	}
 
+	if utils.IsPSPEnabled() {
+		if recResult := psp.CheckEMM(rc); recResult.Completed() {
+			return recResult.Output()
+		}
+
+		// if recResult := psp.CheckPVCHealth(rc); recResult.Completed() {
+		// 	return recResult.Output()
+		// }
```
johntrimble, line 2306: Disabled temporarily since we don't support health timestamp here yet and it made testing EMM easier. We have a ticket to add support for the health timestamp, at which point we will re-enable this.

operator/pkg/utils/k8s_utils.go
```diff
@@ -0,0 +1,160 @@
...
+		if taint.Key == taintKey && taint.Effect == effect {
+			if taint.Value == value {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func FilterNodesWithFn(nodes []*corev1.Node, fn func(*corev1.Node) bool) []*corev1.Node {
```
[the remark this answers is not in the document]
johntrimble: This would have been so much easier in clojure.

sandoichi (approved):

operator/pkg/reconciliation/check_nodes.go
```diff
@@ -25,208 +27,189 @@ func (rc *ReconciliationContext) GetPVCForPod(podNamespace string, podName strin
...
+		if err != nil {
+			return nil, err
+		}
+		pvcs = append(pvcs, pvc)
+	}
+	return pvcs, nil
+}
 
-	wg.Add(1)
+func (rc *ReconciliationContext) getNodesForNameSet(nodeNameSet map[string]bool) ([]*corev1.Node, error) {
```
respringer, line 42: Let's not do this now, but in the future, we should probably define a custom type for NameString of map[string]bool

jimdickinson: map[string]bool is StringSet though?

operator/pkg/psp/emm.go
```diff
@@ -0,0 +1,711 @@
...
+	// EMM operation as cluster availability is already compromised.
+	if len(racksWithDownPods) > 1 {
+		allTaintedNameSet := utils.UnionStringSet(plannedDownNodeNameSet, evacuateDataNodeNameSet)
+		didUpdate := false
+		for nodeName, _ := range allTaintedNameSet {
+			didUpdatePods, err := provider.failEMM(nodeName, TooManyExistingFailures)
+			if err != nil {
+				return result.Error(err)
+			}
+			logger.Info("Failed EMM operation as availability is already compromised due to multiple racks having not ready pods", "node", nodeName)
```
respringer, line 505: I think I'd prefer this at Warn, if this logger has that.  Otherwise error would be best.  I would not want to be a user an have this sort of thing pass by as Info.

operator/pkg/psp/emm.go
```diff
@@ -0,0 +1,711 @@
...
+			nodeNameSetForDownRack)
+
+		if len(nodeNameSetForNoPodsInRack) > 0 {
+			didUpdate := false
+			for nodeName, _ := range nodeNameSetForNoPodsInRack {
+				podsUpdated, err := provider.failEMM(nodeName, TooManyExistingFailures)
+				if err != nil {
+					return result.Error(err)
+				}
+				logger.Info("Failed EMM operation as it did not contain any pods for the currently down rack", "node", nodeName)
```
respringer, line 539: Once again, I'd suggest a Warn/Error here.

respringer (approved):
Left a logging level suggestion. Looks good to me, but there's a metric ton of logic here.  I don't think we will really prove the absence of bugs outside of testing.

operator/pkg/psp/emm.go
jimdickinson: We only have Info and Error, and Error needs an actual non nil err to log with... I think what's in the PR already is basically a warn

operator/pkg/reconciliation/check_nodes.go
johntrimble: Oi! Sorry, missed the ones in this file. Will fix.

operator/pkg/psp/emm.go
respringer: We could use fmt.Errorf(msg) to generate an error from a string:
 
https://golang.org/pkg/fmt/#Errorf 

I am a bit uncomfortable about recording EMM failures as Info, because we are dealing with things that can lead to availability and data failure scenarios.

johntrimble: This case only comes about when their are pods down in multiple racks already and a user sits down, says "frak it", and tries to put a k8s node in EMM anyway. I'd expect there to already be errors logged concerning the pods that are not ready. It is not an error for EMM to fail when availability is compromised. It _is_ an error for availability to be compromised for which I'd expect we already have an error log.

jimdickinson (approved):
thanks for all the comments you put into this one - they helped a lot

PR #290, opened 2020-10-20
Configure stuck pod timeout

sandoichi:
Support configuration of the timeout in which we will delete a stuck pod. 

Setting to 0 will disable the stuck check. 10 minutes is the default (same as current behavior)

### DIFF
--- a/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -146,7 +146,7 @@ type CassandraDatacenterSpec struct {
 	// to false once the restart is in progress.
 	RollingRestartRequested bool `json:"rollingRestartRequested,omitempty"`
 
-	// A map of label keys and values to restrict Cassandra node scheduling to k8s worker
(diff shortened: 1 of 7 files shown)


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -169,6 +169,12 @@ type CassandraDatacenterSpec struct {
 	AdditionalSeeds []string `json:"additionalSeeds,omitempty"`
 
 	Reaper *ReaperConfig `json:"reaper,omitempty"`
+
+	// Determines how many minutes a pod is allowed to exist after
+	// losing readiness before it will automatically be terminated
+	// and replaced.
+	// Default is 10 minutes. Set to 0 to disable
+	TerminateStuckPodsAfterMinutes *int `json:"terminateStuckPodsAfterMinutes,omitempty"`
```
sandoichi, line 177: Open to naming suggestions if this is too verbose

jimdickinson: maybe... `terminationDelayForFailedReadinessPods` and we'll just document that the unit is minutes?

PR #303, opened 2020-10-26
make sure the stopped condition matches the spec

sandoichi:


johntrimble (approved):
Thanks for fixing this!

PR #306, opened 2020-10-30
Allow pod dns lookups for all pods

rchernobelskiy:


rchernobelskiy:
https://github.com/riptano/cloud-ondemand/pull/2280

Issue #309, opened 2020-11-03
The operator can generate incorrect volume mounts 


jsanda:
Note that this issue surface while I was working with a recent build from master, not from a previously released version.

There are a couple problems that can happen you have a `.spec.podTemplateSpec` that includes an initContainer with volume mounts. Here is an example manifest with which I reproduced the issues:

```yaml
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: dc1
spec:
  clusterName: medusa-test
  serverType: cassandra
  serverVersion: 3.11.7
  managementApiAuth:
    insecure: {}
  size: 3
  allowMultipleNodesPerWorker: true
  config:
    jvm-options:
      initial_heap_size: "1024m"
      max_heap_size: "1024m"
  resources:
    requests:
      cpu: 1
      memory: 2Gi
    limits:
      cpu: 1
      memory: 2Gi
  podTemplateSpec:
    spec:
      initContainers:
        - name: medusa-restore
          image: jsanda/medusa:grpc-dev
          imagePullPolicy: IfNotPresent
          env:
            - name: MEDUSA_MODE
              value: RESTORE
          volumeMounts:
            - name: medusa-gcp-key
              mountPath: /etc/medusa-secrets
              subPath: keys
            - name: medusa-config
              mountPath: /etc/medusa/medusa.ini
              subPath: medusa.ini
            - name: server-config
              mountPath: /etc/cassandra
            - mountPath: /var/lib/cassandra
              name: server-data
      containers:
        - name: cassandra
          env:
          - name: JVM_EXTRA_OPTS
            value: -javaagent:/etc/cassandra/jolokia-jvm-1.6.2-agent.jar=port=7373,host=localhost
      volumes:
        - name: medusa-gcp-key
          secret:
            secretName: medusa-gcp-key
        - name: medusa-config
          configMap:
            name: medusa-config
            items:
            - key: medusa.ini
              path: medusa.ini
  storageConfig:
    cassandraDataVolumeClaimSpec:
      storageClassName: standard
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi
```
I declare two additional volumes, `medusa-gcp-key` and `medusa-config`. The `medusa-restore` initContainer declares volume mounts for them as well as for `server-config` and `server-data`. The operator creates the latter two.

Here is the relevant part of the stateful spec that the operator generates:

```yaml
      containers:
      - env:
[...]

PR #310, opened 2020-11-03
do not add volumeMounts from initContainers to cassandra container

jsanda:
This is for #309.

I tested the change and it worked as expected. My change probably broke some tests, so I need to update them if we want to move forward with this.


jsanda:
This is replaced by #336

PR #320, opened 2020-11-24
Use labels instead of zone

### LINKED ISSUE #292: Add the ability to set labels per rack
In order to place pods of each rack in different locations, we need to be able to assign specific labels at the rack level. See https://orange-opensource.github.io/casskop/docs/5_operations/0_implementation_architecture#naming-convention-of-created-objects for an example. It's similar to zones but provides a better granularity.

cscetbon:
Add Labels maps to CassandraDatacenterSpec and replace zone in Rack by Labels too. Now Labels can be set at the DC level and at the rack level and they get merged.

Fixes #292


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -229,8 +232,8 @@ type Rack struct {
 	// The rack name
 	// +kubebuilder:validation:MinLength=2
 	Name string `json:"name"`
-	// Zone name to pin the rack, using node affinity
-	Zone string `json:"zone,omitempty"`
```
jimdickinson, line 233: I would suggest naming the labels field `nodeAffinityLabels`, and we can't just delete the `zone` field here - because backwards compatibility

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -91,6 +91,9 @@ type CassandraDatacenterSpec struct {
 	// Config for the Management API certificates
 	ManagementApiAuth ManagementApiAuthConfig `json:"managementApiAuth,omitempty"`
 
+	//Labels to pin the Datacenter, using node affinity
+	Labels map[string]string `json:"labels,omitempty"`
```
jimdickinson, line 95: `nodeAffinityLabels` here too?

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
cscetbon: @jimdickinson yeah I had it in mind. I can take Zone and add it in the nodeAffinityLabels when it's defined. I could probably also add a comment to mark **Zone** as deprecated

operator/pkg/reconciliation/utils.go
```diff
@@ -36,4 +36,11 @@ func getResourcesOrDefault(res *corev1.ResourceRequirements,
 	}
 
 	return res
-}
\ No newline at end of file
+}
+
+func emptyMapIfNil(m map[string]string) map[string]string {
```
jimdickinson, line 41: 👍

operator/pkg/reconciliation/construct_statefulset.go
```diff
@@ -80,6 +79,29 @@ func shouldDefineSecurityContext(dc *api.CassandraDatacenter) bool {
...
+			nodeAffinityLabels = utils.MergeMap(emptyMapIfNil(rack.NodeAffinityLabels),
+				emptyMapIfNil(dc.Spec.NodeAffinityLabels))
+			if rack.Zone != "" {
+				if _, found := nodeAffinityLabels["zone"]; found {
+					err := fmt.Errorf("Deprecated parameter Zone is used and also defined in " +
+						"NodeAffinityLabels. You should only define it in NodeAffinityLabels")
+					return nil, err
+				}
+				nodeAffinityLabels = utils.MergeMap(
+					emptyMapIfNil(nodeAffinityLabels), map[string]string{"zone": rack.Zone},
```
jimdickinson, line 96: should be 
```suggestion
					emptyMapIfNil(nodeAffinityLabels), map[string]string{"failure-domain.beta.kubernetes.io/zone": rack.Zone},
```

I think?

operator/pkg/reconciliation/construct_statefulset.go
```diff
@@ -80,6 +79,29 @@ func shouldDefineSecurityContext(dc *api.CassandraDatacenter) bool {
 	return dc.Spec.ServerType == "dse" || images.CalculateDockerImageRunsAsCassandra(dc.Spec.ServerVersion)
 }
 
+func rackNodeAffinitylabels(dc *api.CassandraDatacenter, rackName string) (nodeAffinityLabels map[string]string,
+	Error error) {
```
jimdickinson, line 83: we have favored not using named return values

operator/pkg/reconciliation/construct_statefulset.go
```diff
@@ -80,6 +79,29 @@ func shouldDefineSecurityContext(dc *api.CassandraDatacenter) bool {
...
+	racks := dc.GetRacks()
+	for _, rack := range racks {
+		if rack.Name == rackName {
+			nodeAffinityLabels = utils.MergeMap(emptyMapIfNil(rack.NodeAffinityLabels),
+				emptyMapIfNil(dc.Spec.NodeAffinityLabels))
+			if rack.Zone != "" {
+				if _, found := nodeAffinityLabels["zone"]; found {
```
jimdickinson, line 90: I think this key would also change?

operator/pkg/reconciliation/construct_statefulset.go
```diff
@@ -80,6 +79,29 @@ func shouldDefineSecurityContext(dc *api.CassandraDatacenter) bool {
...
+					err := fmt.Errorf("Deprecated parameter Zone is used and also defined in " +
+						"NodeAffinityLabels. You should only define it in NodeAffinityLabels")
+					return nil, err
```
jimdickinson, line 93: I think we can just document that the labels override the zone field, but I think logging the above message is a good idea too

cscetbon: Okay so logging but not failing, correct ?

operator/pkg/reconciliation/construct_statefulset.go
cscetbon: it avoids to define them but as you wish

operator/pkg/reconciliation/construct_statefulset.go
jimdickinson: yeah I think so

cscetbon:
@jimdickinson I made the different changes that were asked. I'm not sure about the changes to make at the documentation or  examples level. Any guidance ?

operator/pkg/reconciliation/construct_statefulset.go
```diff
@@ -7,20 +7,22 @@ package reconciliation
...
+	"github.com/datastax/cass-operator/operator/pkg/utils"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
+const ZONE_LABEL = "failure-domain.beta.kubernetes.io/zone"
```
jimdickinson, line 24: the linter will complain if there's all caps or underscores, and we've been following that 🤷

```suggestion
const ZoneLabel = "failure-domain.beta.kubernetes.io/zone"
```

sorry!

cscetbon: No worries my bad, switching between Python and Go at this time lol. Will fix it in the next 10 min 😉

jimdickinson:
> @jimdickinson I made the different changes that were asked. I'm not sure about the changes to make at the documentation or examples level. Any guidance ?

First of all - this PR is looking really good - thank you!

If you'd like to take a stab at docs, looks like we have some vague ones here -> https://github.com/datastax/cass-operator/blob/master/docs/user/README.md#racks

The examples are more clear, but terse -> https://github.com/datastax/cass-operator/blob/master/operator/example-cassdc-yaml/cassandra-3.11.x/example-cassdc-full.yaml#L27

respringer (approved):
Looks good to me.  Approved.

johntrimble (approved):
Looks great!

operator/example-cassdc-yaml/dse-6.8.x/example-cassdc-full.yaml
```diff
@@ -24,11 +24,11 @@ spec:
...
   racks:
     - name: rack1
-      zone: us-central1-a
+      failure-domain.beta.kubernetes.io/zone: us-central1-a
     - name: rack2
-      zone: us-central1-b
+      failure-domain.beta.kubernetes.io/zone: us-central1-b
     - name: rack3
-      zone: us-central1-c
+      failure-domain.beta.kubernetes.io/zone: us-central1-c
```
jimdickinson, line 31: just noticed this needs the
```
      nodeAffinityLabels:
```
layer

jimdickinson (approved):

jimdickinson (approved):

mars64:
Thanks for this! Any ETA when this will land in a release?

PR #330, opened 2020-12-08
Additional storage volumes for sidecars or types of cassandra files

### LINKED ISSUE #327: Allow to add PVCs in order to retain data or provide some room to sidecars
As a more advanced users feature and to be CassKop compatible, I'm proposing to add this feature which would provide the ability to add PVCs to : 
- provide more space to retain logs (let's imagine cdc is enabled and it needs more room)
- provide other types of PVCs for certain types of files
...

In CassKop the [same parameter](https://github.com/Orange-OpenSource/casskop/blob/v1.0.1-release/pkg/apis/db/v1alpha1/cassandracluster_types.go#L877-L886) is used but it hosts PVCs that can then be referred to in sidecarConfigs. The controller then [create those PVCs](https://github.com/Orange-OpenSource/casskop/blob/v1.0.1-release/pkg/controller/cassandracluster/generator.go#L223-L235).

I'd like to propose the same here and then those PVCs can be used in podTemplateSpec. If we go down that road we should probably use a different name to specify those PVCs in order to be backward compatible for existing users.

cscetbon:
This PR fixes #327 by allowing to add PVCs to each statefulset created. It would be typically used by advanced users.


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -208,8 +208,22 @@ type DseWorkloads struct {
...
 
+// StorageConfig defines additional storage configurations
+type AdditionalVolumes struct {
+	// Mount path into cassandra container
+	MountPath string `json:"mountPath"`
+	// Name of the pvc
+	// +kubebuilder:validation:Pattern=[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*
+	Name string `json:"name"`
+	// Persistent volume claim spec
+	PVCSpec *corev1.PersistentVolumeClaimSpec `json:"pvcSpec"`
```
jimdickinson, line 219: not sure if the codegen rules would prefer this to be `PvcSpec`... I'm going to find them

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -208,8 +208,22 @@ type DseWorkloads struct {
 	SearchEnabled    bool `json:"searchEnabled,omitempty"`
 }
 
+// StorageConfig defines additional storage configurations
+type AdditionalVolumes struct {
+	// Mount path into cassandra container
+	MountPath string `json:"mountPath"`
+	// Name of the pvc
+	// +kubebuilder:validation:Pattern=[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*
```
jimdickinson, line 216: could you explain this pattern in a comment? I don't understand the dot and later part

cscetbon: @jimdickinson it's there to validate that we enter a name that can be used to create a resource. I took that one from casskop's code. the first part is supposed to allow to specify a namespace but it's wrong cause PersistentVolumeClaim is a namespaced object. Removing that part. The later part just prevents other characters like colon, underscores etc... It's supposed to avoid creating part of the resources and failing

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
cscetbon: I expect it to accept it cause that's what is used in the API, for instance see https://godoc.org/k8s.io/kubernetes/pkg/api/persistentvolumeclaim#DropDisabledFields, they don't write it oldPvcSpec

jsanda:
@cscetbon In [k8ssandra/k8ssandra#66](https://github.com/k8ssandra/k8ssandra/pull/66) you gave the following example:

```yaml
 storageConfig:
    cassandraDataVolumeClaimSpec: &rw_1g_vol
      storageClassName: server-storage
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
    additionalVolumes:
      - name: server-logs
        mountPath: /var/log/cassandra
        pvcSpec: *rw_1g_vol
      - name: cassandra-commitlogs
        mountPath: /var/lib/cassandra/commitlog
        pvcSpec: *rw_1g_vol
```
How would the `cassandra-commitlogs` volume mount work? Would it shadow mount `/var/lib/cassandra`?

cscetbon:
@jsanda if it's a known volume (like server-logs) from the cassandra pod then it would be used as is. If it's not then it needs to be mounted in the cassandra container as well and I think you would just update podTemplateSpec to do so.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
jimdickinson: nothing to see here, I think older versions of the code generation tools had more requirements

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -211,8 +211,22 @@ type DseWorkloads struct {
...
 
+// StorageConfig defines additional storage configurations
+type AdditionalVolumes struct {
+	// Mount path into cassandra container
+	MountPath string `json:"mountPath"`
+	// Name of the pvc
+	// +kubebuilder:validation:Pattern=[a-z0-9]([-a-z0-9]*[a-z0-9])?
+	Name string `json:"name"`
+	// Persistent volume claim spec
+	PVCSpec *corev1.PersistentVolumeClaimSpec `json:"pvcSpec"`
```
burmanm, line 222: Does it make sense to keep this as a pointer? Each time it is used in the code the pointer is de-referenced.

burmanm:
What happens when the DC is deleted? Are the PVCs meant to stay behind or should they be deleted also?

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
cscetbon: @burmanm I did the same as what is done for field `CassandraDataVolumeClaimSpec`, but that field can be missing while pvcSpec is really required. It's no longer a pointer in my last commit

cscetbon:
> What happens when the DC is deleted? Are the PVCs meant to stay behind or should they be deleted also?

@burmanm  PVCs are part of a statefulset so they'll be deleted. But for the PV It'll depend on the reclaim policy set in the storage class used.

burmanm:
Deleting a StatefulSet does not delete PVCs, at least according to the issues in Kubernetes and the docs.  https://github.com/kubernetes/kubernetes/issues/55045

Or is this not what you've observed?

jsanda:
cass-operator deletes the PVCs. See [here](http://github.com/datastax/cass-operator/blob/5ef199a74df1916c21c900c65795abbdab1cd05c/operator/pkg/reconciliation/reconcile_datacenter.go#L38-L38).

operator/pkg/reconciliation/construct_podtemplatespec.go
```diff
@@ -373,7 +409,7 @@ func buildContainers(dc *api.CassandraDatacenter, baseTemplate *corev1.PodTempla
 
 	cassContainer.Ports = combinePortSlices(portDefaults, cassContainer.Ports)
 
-	// Combine volumes
+	// Combine volumeMounts
```
jimdickinson, line 412: 👍

tests/additional_volumes/additional_volumes_suite_test.go
```diff
@@ -0,0 +1,70 @@
+// Copyright DataStax, Inc.
+// Please see the included license file for details.
+
+package scale_up
```
jimdickinson, line 4:
```suggestion
package additional_volumes
```

jimdickinson (approved):

PR #335, opened 2020-12-17
remove Reaper integration

jsanda:
Removing Reaper integration because Reaper sidecar mode had too many problems that required changes in Reaper to make it work better. See https://github.com/thelastpickle/cassandra-reaper/issues/956 for details about Reaper sidecar issues. 

I instead decided to implement Reaper support through [k8ssandra](http://k8ssandra.io)


operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -174,8 +174,6 @@ type CassandraDatacenterSpec struct {
 
 	AdditionalSeeds []string `json:"additionalSeeds,omitempty"`
 
-	Reaper *ReaperConfig `json:"reaper,omitempty"`
```
jimdickinson, line 177: what happens if I had this and I was using 1.5.0, then I upgrade and we rip it out?

jsanda: That's a good question. I will test and report back my findings.

PR #336, opened 2020-12-17
Make sure correct volume mounts are created

### LINKED ISSUE #309: The operator can generate incorrect volume mounts

jsanda:
Fixes #309

### DIFF
--- a/operator/pkg/reconciliation/construct_podtemplatespec_test.go
+++ b/operator/pkg/reconciliation/construct_podtemplatespec_test.go
@@ -114,10 +114,10 @@ func TestCassandraDatacenter_buildInitContainer_with_overrides(t *testing.T) {
 	podTemplateSpec := &corev1.PodTemplateSpec{
 		Spec: corev1.PodSpec{
 			InitContainers: []corev1.Container{
-				corev1.Container{
+				{
 					Name: ServerConfigContainerName,
 					Env: []corev1.EnvVar{
-						corev1.EnvVar{
+						{
 							Name:  "k1",
 							Value: "v1",
 						},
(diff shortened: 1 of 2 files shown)


operator/pkg/reconciliation/construct_podtemplatespec_test.go
```diff
@@ -410,6 +410,280 @@ func TestCassandraDatacenter_buildPodTemplateSpec_initcontainers_merge(t *testin
 	}
 }
 
+func TestCassandraDatacenter_buildPodTemplateSpec_add_initContainer_after_config_initContainer(t *testing.T) {
+	// When adding an initContainer with podTemplate spec it will run before
+	// the server config initContainer by default. This test demonstrates and
+	// verifies how to specify the initContainer to run after the server config
+	// initContainer.
```
jimdickinson, line 417: 💯

operator/pkg/reconciliation/construct_podtemplatespec_test.go
```diff
@@ -410,6 +410,280 @@ func TestCassandraDatacenter_buildPodTemplateSpec_initcontainers_merge(t *testin
...
+
+type VolumeMountMatcher func(volumeMount corev1.VolumeMount) bool
+
+type VolumeMatcher func(volume corev1.Volume) bool
+
+func volumeNameMatcher(name string) VolumeMatcher {
+	return func(volume corev1.Volume) bool {
+		return volume.Name == name
+	}
+}
```
jimdickinson, line 652: I'm good with this, but I feel obliged to point out Rob Pike's repo here -> https://github.com/robpike/filter

operator/pkg/reconciliation/construct_podtemplatespec_test.go
```diff
@@ -436,7 +710,7 @@ func TestCassandraDatacenter_buildPodTemplateSpec_labels_merge(t *testing.T) {
 }
 
 // A volume added to ServerConfigContainerName should be added to all built-in containers
-func TestCassandraDatacenter_buildPodTemplateSpec_propagate_volumes(t *testing.T) {
+func TestCassandraDatacenter_buildPodTemplateSpec_do_not_propagate_volumes(t *testing.T) {
```
jimdickinson, line 713: 🤣

jimdickinson (approved):
lgtm - thank you for the great tests and comments!

operator/pkg/reconciliation/construct_podtemplatespec_test.go
jsanda: Do you want me to update the code to use that? Happy to do so. I promise you I do not have NIH syndrome :)

jimdickinson: to get rid of all the one line functions and functions abstracting a for loop? no it reads cleanly here, and this is test code, if it's working right now let's keep it as-is

jimdickinson (approved):
lgtm

PR #311, opened 2020-11-09
Allow configuration of the system.log tail-er

jimdickinson:
- Start adding configuration for logging sidecar
- Implement the logic for both new configs
- Update docstrings
- Adding unit tests

fixes #241

PR #339, opened 2020-12-17
operator/pkg/images: Allow busybox image to be set by the user

nesv:
Thank you for taking the time to look at this pull request!

In some Kubernetes environments, the operators may opt to set up [ImagePolicyWebhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#imagepolicywebhook) or any other validating admission webhook (or admission controller) to restrict the registry or tags of the images that are deployed. If an image tag is not specified, it defaults to `latest`. In my case, we have an admission webhook which prevents us from deploying any images' `latest` tag.

In this pull request, the default Busybox image is still `busybox`, but the user can set the `BUSYBOX_IMAGE` environment variable to either lock the busybox image to a specific version, and/or use a different image registry.

### DIFF
--- a/operator/pkg/images/images.go
+++ b/operator/pkg/images/images.go
@@ -20,6 +20,7 @@ const (
 	envDefaultRegistryOverride            = "DEFAULT_CONTAINER_REGISTRY_OVERRIDE"
 	envDefaultRegistryOverridePullSecrets = "DEFAULT_CONTAINER_REGISTRY_OVERRIDE_PULL_SECRETS"
 	EnvBaseImageOS                        = "BASE_IMAGE_OS"
+	EnvBusyboxImage                       = "BUSYBOX_IMAGE"
 	ValidDseVersionRegexp                 = "6\\.8\\.\\d+"
 	ValidOssVersionRegexp                 = "(3\\.11\\.\\d+)|(4\\.0\\.\\d+)"
 	UbiImageSuffix                        = "-ubi7"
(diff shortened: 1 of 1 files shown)


jimdickinson:
Hi @nesv - did you see in 1.5.0 the log tailer did become customizable. I do see the use of not picking `latest` as a default... Do you want to change this PR to add a tag to `busybox`. Looks like the equivalent behavior based on [this page](https://hub.docker.com/_/busybox?tab=tags&page=1&ordering=last_updated) would be `busybox:1.32.0-uclibc`. I think it's fine for this to become the norm - no env vars needed.
`

nesv:
Hi @jimdickinson!

> did you see in 1.5.0 the log tailer did become customizable? See #311.

My motivation behind this pull request doesn't have anything to do with the log tailer; it's only about being able to override the busybox image so that it can work in environments that either prohibit the use of Docker Hub (or, inversely, require the use of a private/internal registry) or prohibit the use of the `latest` tag on images.

> Do you want to change this PR to add a tag to busybox?

I can't seem to add any labels to this PR. Would you like me to close this one, and re-open it with the "busybox" label?

jimdickinson:
Right - I'm trying to say that since the linked PR went in, you can specify any image location for the logging container at runtime, per each `CassandraDatacenter`. I do think avoiding `latest` is a great practice, though, so I'd like to get that in. Can you update this PR?

nesv:
> you can specify any image location for the logging container at runtime

Gotcha. I'm sorry, I didn't realize `busybox` was used as the logging container. I'll admit, I was a little hyper-focused on just being able to change the image.

> I do think avoiding latest is a great practice, though, so I'd like to get that in. Can you update this PR?

Happily. :slightly_smiling_face:

nesv:
@jimdickinson I've updated the branch to set the tag for the "busybox" image.

jimdickinson:
ack - sorry I lost track of this PR. Would like to merge it - would you mind going through the flow at cla.datastax.com first @nesv ?

nesv:
@jimdickinson Done!

jimdickinson (approved):

jimdickinson:
Thank you!

nesv:
@jimdickinson You're most welcome!

PR #351, opened 2021-01-06
use pre-upgrade hook to set resource policy on CRD

### LINKED ISSUE #324: Multiple installs in namespace mode fail
Currently if we try to install the operator in 2 different namespaces using helm, the second install fails with 
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition "cassandradatacenters.cassandra.datastax.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "cass-operator-1606499002"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "cass-operator-1606498846"
```

The reason is that CRDs are cluster-wide and can't be installed twice. But your helm chart install it no matter what.

helm 3 supports now a folder called crds, see https://helm.sh/docs/chart_best_practices/custom_resource_definitions, and would skip that step if they already exist. This is what should be used instead.

Here are the steps I followed:
```
$ kubectl create namespace casskop-install1
namespace/casskop-install1 created
$ kubectl create namespace casskop-install2
namespace/casskop-install2 created
$ helm install --generate-name --set clusterWideInstall=false -n casskop-install1 datastax/cass-operator 

NAME: cass-operator-1606500664
LAST DEPLOYED: Fri Nov 27 13:11:07 2020
NAMESPACE: casskop-install1
STATUS: deployed
REVISION: 1
TEST SUITE: None
$ helm install --generate-name --set clusterWideInstall=false -n casskop-install2 datastax/cass-operator 

Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition 
"cassandradatacenters.cassandra.datastax.com" in namespace "" exists and cannot be imported into the current release: 
invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal 
"cass-operator-1606500712": current value is "cass-operator-1606500664"; annotation validation error: key 
"meta.helm.sh/release-namespace" must equal "casskop-install2": current value is "casskop-install1"
```

jsanda:
Fixes #324

In #304 customresourcedefinition.yaml was moved back to the templates directory because on `helm upgrade` Helm would delete the CRD since it no longer exists in the templates directory. This has come up a number of times in the Helm community. See https://github.com/helm/helm/issues/8163 for background.

In this PR I move customresourcedefinition.yaml back to the crds directory. I use a pre-upgrade hook to add an annotation to the CRD that tells Helm to preserve the CRD even though it has been moved. I tested this manually. I also ran `M_INTEG_DIR=upgrade_operator mage integ:run` to make sure the upgrade works.


cscetbon:
@jsanda @jimdickinson any news on that issue ?

PR #356, opened 2021-01-13
retrying decommission instead of assuming it works

rsds143:
We were hitting a corner where if the decom never succesfully regsitered on the node it would never try again and the cluster would be stuck in a decom state that could never come.

highlights
- now check for to see if the decommission actually started and then retrying the decommission if it has not
- now logging decommission service errors. this is probably needless but once the mgmt api in dse gets more solid and can handle returning a "started startus" and we can do some proper error handling the logging maybe useful.

NOTE: labeling this as a wip to get some tests running but this appears to work

### DIFF
--- a/operator/pkg/reconciliation/decommission_node.go
+++ b/operator/pkg/reconciliation/decommission_node.go
@@ -127,9 +127,8 @@ func (rc *ReconciliationContext) DecommissionNodeOnRack(rackName string, epData 
 			}
 
 			if err := rc.NodeMgmtClient.CallDecommissionNodeEndpoint(pod); err != nil {
-				// TODO this returns a 500 when it works
-				// We are waiting for a new version of mgmt api with a fix for this
-				// return err
+				rc.ReqLogger.Info(fmt.Sprintf("Error from decommission attempt. This is only an attempt and can"+
+					" fail it will be retried later if decomission has not started. Error: %v", err))
 			}
 
 			rc.ReqLogger.Info("Marking node as decommissioning")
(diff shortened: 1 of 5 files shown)


operator/pkg/reconciliation/decommission_node.go
```diff
@@ -158,7 +157,12 @@ func (rc *ReconciliationContext) CheckDecommissioningNodes(epData httphelper.Cas
 
 	for _, pod := range rc.dcPods {
 		if pod.Labels[api.CassNodeState] == stateDecommissioning {
-			if !IsDoneDecommissioning(pod, epData) {
+			if !HasStartedDecommissioning(pod, epData) {
```
sandoichi, line 160: Is there a risk here of having it get stuck whenever a pod finishes decommissioning? Like, if the pod status is `LEFT`, we will always enter this first if block right?

should it be changed to:
```
if !HasStartedDecommissioning... && !IsDoneDecommissioning...
```
or move the new block inside of the old one like:
``` 
if !IsDoneDecommissioning(pod, epData) {
   if !HasStartedDecommissioning....
```
instead?

rsds143: the name is maybe bad but no they are looking for mutually exclusive statuses 

func HasStartedDecommissioning(pod *v1.Pod, epData httphelper.CassMetadataEndpoints) bool {
	for idx := range epData.Entity {
		ep := &epData.Entity[idx]
		if ep.GetRpcAddress() == pod.Status.PodIP {
			return !strings.HasPrefix(ep.Status, "LEAVING")
		}
	}
	return true
}

func IsDoneDecommissioning(pod *v1.Pod, epData httphelper.CassMetadataEndpoints) bool {
	for idx := range epData.Entity {
		ep := &epData.Entity[idx]
		if ep.GetRpcAddress() == pod.Status.PodIP {
			return strings.HasPrefix(ep.Status, "LEFT")
		}
	}

rsds143:
Ok I feel pretty good where this is at, stopping work on it unless there is more than you would like added

operator/pkg/reconciliation/decommission_node_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/datastax/cass-operator/operator/internal/result"
+	api "github.com/datastax/cass-operator/operator/pkg/apis/cassandra/v1beta1"
+	"github.com/datastax/cass-operator/operator/pkg/httphelper"
+	"github.com/datastax/cass-operator/operator/pkg/mocks"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestRetryDecommissionNode(t *testing.T) {
```
jimdickinson, line 27: Really nice usage of the mocks!

jimdickinson (approved):

operator/pkg/reconciliation/decommission_node.go
jimdickinson: nice catch

PR #389, opened 2021-02-26
Reduce dependencies of apis/CassandraDatacenter to allow newer Kubernetes

burmanm:
When using cass-operator's CassandraDatacenter as a dependency in another project, the current implementation prevents the use of newer versions of Kubernetes dependencies (conflicts with certain removed APIs).

These dependencies come from the use of utils package as well as images. Images pulls old controller-runtime which wants old non-existant Kube APIs and the utils picks up also persistentvolume dependencies.

I think the end approach should be that apis does not call other packages inside cass-operator to ensure that it's self contained and can be more easily used in other projects as dependency. However, with these modifications I was able to use CassandraDatacenter as a dependency with Kubernetes versions up to 1.20.4 (newest released), so it solves the problem. serverconfig does not use external packages, but there was a note of code-gen requirement to keep it in the external package - which probably requires updating the controller-runtime to operator-sdk 1.x series - and that's a lot bigger project.

### DIFF
--- a/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
+++ b/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
@@ -8,14 +8,11 @@ import (
 	"fmt"
 
 	"github.com/Jeffail/gabs"
+	"github.com/datastax/cass-operator/operator/pkg/serverconfig"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-
-	"github.com/datastax/cass-operator/operator/pkg/images"
-	"github.com/datastax/cass-operator/operator/pkg/serverconfig"
-	"github.com/datastax/cass-operator/operator/pkg/utils"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


jsanda:
>I think the end approach should be that apis does not call other packages inside cass-operator to ensure that it's self contained and can be more easily used in other projects as dependency.

For reference we pull in cass-operator as a dependency in the following projects:

* https://github.com/k8ssandra/k8ssandra
* https://github.com/k8ssandra/reaper-operator
* https://github.com/k8ssandra/medusa-operator

jsanda:
I manually tested in k8ssandra. LGTM!

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -8,14 +8,11 @@ import (
...
 	"github.com/Jeffail/gabs"
+	"github.com/datastax/cass-operator/operator/pkg/serverconfig"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-
-	"github.com/datastax/cass-operator/operator/pkg/images"
-	"github.com/datastax/cass-operator/operator/pkg/serverconfig"
-	"github.com/datastax/cass-operator/operator/pkg/utils"
```
jimdickinson, line 18: I was confused at first - but the important thing is to get as few dependencies from this v1beta1 package as possible right? That's something I noticed some time ago, so I agree with this 👍

jimdickinson (approved):

burmanm:
Available in v1.7.0 in k8ssandra/cass-operator

PR #392, opened 2021-03-02
Update default version to fix CVEs

zencircle:
https://github.com/datastax/management-api-for-apache-cassandra/issues/83

### DIFF
--- a/operator/pkg/images/images.go
+++ b/operator/pkg/images/images.go
@@ -77,7 +77,7 @@ const (
 var imageLookupMap map[Image]string = map[Image]string{
 
 	Cassandra_3_11_6: "datastax/cassandra-mgmtapi-3_11_6:v0.1.5",
-	Cassandra_3_11_7: "datastax/cassandra-mgmtapi-3_11_7:v0.1.13",
+	Cassandra_3_11_7: "datastax/cassandra-mgmtapi-3_11_7:v0.1.22",
 	Cassandra_4_0_0:  "datastax/cassandra-mgmtapi-4_0_0:v0.1.12",
 
 	UBICassandra_3_11_6: "datastax/cassandra:3.11.6-ubi7",


operator/pkg/images/images.go
```diff
@@ -77,7 +77,7 @@ const (
 var imageLookupMap map[Image]string = map[Image]string{
 
 	Cassandra_3_11_6: "datastax/cassandra-mgmtapi-3_11_6:v0.1.5",
-	Cassandra_3_11_7: "datastax/cassandra-mgmtapi-3_11_7:v0.1.13",
+	Cassandra_3_11_7: "datastax/cassandra-mgmtapi-3_11_7:v0.1.22",
```
jsanda, line 80: Thanks for the PR @zencircle.

The `v0.1.22` image is configured to run Cassandra as non-root whereas the `v0.1.13` image is configured to run as root. While running as non-root is definitely a good change, switching the default image in this way probably merits some discussion.

emerkle826:
@zencircle You will also need to update the test here:
https://github.com/zencircle/cass-operator/blob/patch-1/operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types_test.go#L45
to change the image version as well.

operator/pkg/images/images.go
emerkle826: @zencirle, John is right. You will likely run into problems with this bit of code:

https://github.com/zencircle/cass-operator/blob/patch-1/operator/pkg/images/images.go#L168-L179

ss it will assume that the default 3.11.7 image is run as root. The version you are upgrading too does not run as root as John said.

emerkle826: @zencircle
You likely have to remove "3.11.7" from this line:
https://github.com/zencircle/cass-operator/blob/patch-1/operator/pkg/images/images.go#L172

and change the line in the test that expects it to be run as root (from false to true) here:
https://github.com/zencircle/cass-operator/blob/patch-1/operator/pkg/images/images_test.go#L70

zencircle:
@emerkle826 I fixed all the issues, now its failing on ecr credentials

emerkle826:
> @emerkle826 I fixed all the issues, now its failing on ecr credentials

I'm not 100% sure, but I believe the ECR failure is because you do not have AWS credentials configured. This is likely only going to pass for PRs based on branches in the `datastax` repo.

Also, there is some concern at the moment about changing the default Cassandra imaged used by cass-operator such that the process runs as the `cassandra` user (changes made to the [Management API image](https://github.com/datastax/management-api-for-apache-cassandra/releases/tag/v0.1.20) as of v0.1.20). You can override the default image if you wish, however. Is this not a sufficient workaround?

zencircle:
Workaround is to patch the datacenter, `dc1-patch.yaml`
```
apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
spec:
  serverImage: datastax/cassandra-mgmtapi-3_11_7:v0.1.22
```
```
kubectl patch cassdc dc1 --type merge -p "$(cat dc1-patch.yaml)"
```

PR #398, opened 2021-03-08
Add additionalServiceOptions

### LINKED ISSUE #293: Allow for specifying additional labels on client service
In EKS, when stateful sets have an external DNS annotation applied (`external-dns.alpha.kubernetes.io/hostname`) the underlying pods become available at an external DNS address.

In GKE, this same tactic results in, nothing. Instead, the annotation must be applied to the service (ideally the client service) for the same behavior to be exhibited.

The CassandraDatacenter spec should allow for specifying labels on the services it creates

burmanm:
Allows defining additional labels and/or annotations to generated services. At the moment it only creates it to the main-service (<cluster>-<datacenter>-service) - do we want the same replicated to cluster2-seed-service and cluster2-dc2-all-pods-service as well ?

Fixes #293 and https://github.com/k8ssandra/k8ssandra/issues/478


operator/pkg/reconciliation/construct_service.go
```diff
@@ -65,6 +65,19 @@ func newServiceForCassandraDatacenter(dc *api.CassandraDatacenter) *corev1.Servi
 
 	utils.AddHashAnnotation(service)
 
+	serviceConfig := dc.Spec.AdditionalServiceConfig
+	if serviceConfig.Labels != nil && len(serviceConfig.Labels) > 0 {
```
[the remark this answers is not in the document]
burmanm: So cass-operator labels > user-set labels?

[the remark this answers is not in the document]
jimdickinson: We used to favor that stance, but inevitably there's always some corner case where not accepting a user's override breaks a use case, so we've generally flipped to user data always overrides operator data, at the risk of the breaking things. 🙁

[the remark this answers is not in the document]
jsanda: Thanks for the background. So I guess it is _use at your own risk_ then.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -181,6 +181,9 @@ type CassandraDatacenterSpec struct {
 
 	// Container image for the log tailing sidecar container.
 	SystemLoggerImage string `json:"systemLoggerImage,omitempty"`
+
+	// AdditionalServiceConfig allows to define additional parameters that are included in the created Services
```
jsanda, line 185: Can we add a warning blurb here to let the user know he can break things by overwriting labels set by cass-operator so extra caution should take when setting this property.

jsanda (changes requested):

jsanda (changes requested):
@burmanm not sure if you have seen the latest comments in #293 but the request is to be able to specify labels and annotation per service.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -241,6 +244,21 @@ func (dc *CassandraDatacenter) GetRacks() []Rack {
 	}}
 }
 
+// ServiceConfig defines additional service configurations. Note, user can override values set by cass-operator, doing so could break cass-operator functionality.
+type ServiceConfig struct {
+	DatacenterService     ServiceConfigAdditions `json:"dcService,omitempty"`
```
jsanda, line 249: Should these be pointers? I ask because I wonder if it would make it easier to determine if the user sets the property, particularly when dealing with nested fields.

burmanm: No, it works fine as it is. I would still need to check the separately the Labels and Annotations.

jsanda (changes requested):
@burmanm The changes look good from my perspective. The only remaining request I have is a minor doc update to indicate that user-supplied labels will overwrite those created by cass-operator.

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
burmanm: Ah, it doesn't copy it from the ServiceConfig description.

jsanda (approved):
Changes LGTM. 

@jimdickinson can you review please?

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
```diff
@@ -181,6 +181,9 @@ type CassandraDatacenterSpec struct {
 
 	// Container image for the log tailing sidecar container.
 	SystemLoggerImage string `json:"systemLoggerImage,omitempty"`
+
+	// AdditionalServiceConfig allows to define additional parameters that are included in the created Services. Note, user can override values set by cass-operator and doing so could break cass-operator functionality.
```
jimdickinson, line 185: 👍 , but we'll probably have to document which labels they should avoid, somewhere

operator/pkg/reconciliation/construct_service.go
```diff
@@ -65,9 +65,24 @@ func newServiceForCassandraDatacenter(dc *api.CassandraDatacenter) *corev1.Servi
 
 	utils.AddHashAnnotation(service)
 
+	addAdditionalOptions(service, &dc.Spec.AdditionalServiceConfig.DatacenterService)
```
jimdickinson, line 68: I think we should put this above the `AddHashAnnotation` call, or it won't roll out an update to the service

operator/pkg/reconciliation/construct_service.go
```diff
@@ -95,6 +110,7 @@ func newSeedServiceForCassandraDatacenter(dc *api.CassandraDatacenter) *corev1.S
 	service.Spec.PublishNotReadyAddresses = true
 
 	utils.AddHashAnnotation(service)
+	addAdditionalOptions(service, &dc.Spec.AdditionalServiceConfig.SeedService)
```
jimdickinson, line 113: same for all of these

jimdickinson (approved):
👍  looks really good! just `addAdditionalOptions` before we `AddHashAnnotation`

operator/pkg/reconciliation/construct_service.go
jsanda: I totally missed that. Thanks 😄

operator/pkg/apis/cassandra/v1beta1/cassandradatacenter_types.go
jsanda: @burmanm can you document the labels that need to be avoided? I think there are only a handful, in which case maybe just list them in the comments here.

jsanda (changes requested):

jsanda (approved):
👍

Issue #382, opened 2021-02-22
CRD for creating keyspaces and roles


michelderu:
In order to increase security it would be great if keyspaces and roles can be created by use of a CRD. This allows for better integration in Infrastructure As Code. And it prevents an organization from opening up the MGMT API to do this.

PR #400, opened 2021-03-09
Add CreateKeyspace as well as AlterKeyspace for httphelper

burmanm:
These methods are needed by https://github.com/k8ssandra/k8ssandra/issues/227 (to avoid replicating the client code in reaper-operator) as well features like https://github.com/datastax/cass-operator/issues/382 (if we want to implement it)


burmanm:
Available in 1.7.0 (k8ssandra/cass-operator)

PR #409, opened 2021-03-26
Override container user Id from env

skhatri:
We have a managed Kubernetes Environment where it is enforced that all uids be between 1000 and 65535.
To fix this, we are using DSE image and are creating a user 1000 so our deployment is not rejected.

The operator hard codes the value to 999. This PR defaults the userID to 999 but will try to use the DSE_USER_ID environment variable if provided.

### DIFF
--- a/operator/pkg/reconciliation/construct_statefulset.go
+++ b/operator/pkg/reconciliation/construct_statefulset.go
@@ -13,7 +13,9 @@ import (
 	"github.com/datastax/cass-operator/operator/pkg/oplabels"
 	"github.com/datastax/cass-operator/operator/pkg/psp"
 	"github.com/datastax/cass-operator/operator/pkg/utils"
+	"os"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"strconv"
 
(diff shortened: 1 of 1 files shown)


jsanda (commented):
Hi @skhatri 

Thanks for the PR! 

Personally I am not a fan of using an environment variable since it does not allow be to configure the SecurityContext per CassandraDatacenter.

It turns out that you can already override the SecurityContext via the `.spec.podTemplateSpec` property. If that is not sufficient for your requirements, I would rather see a SecurityContext property added to the CRD.

skhatri:
The above PR was the equivalent of what we did with 1.4.1. The ```shouldDefineSecurityContext``` was not in 1.4.1
Thanks for your suggestion. Will try that out.

skhatri:
Can confirm that operator 1.6.0 supports this and merges securityContext at container level as expected. 

```
  podTemplateSpec:
    spec:
      containers: 
        - name: cassandra
          securityContext:
            runAsUser: 1000
            runAsNonRoot: true
            readOnlyRootFilesystem: true
```

closing this issue.

Issue #110, opened 2020-06-08
Watch multiple namespaces?


ninjaslothx:
I see the container spec has an env variable for WATCH_NAMESPACE. 
https://github.com/datastax/cass-operator/blob/master/docs/user/cass-operator-manifests-v1.17.yaml#L6308

Is it possible to have the operator watch multiple or all namespaces?

jimdickinson:
Yes, that env var can accept a comma delimited list of namespaces. Clearing it should have one operator watch all namespaces.

source: https://sdk.operatorframework.io/docs/golang/quickstart/#manager

ninjaslothx:
Thank you! Works as described.

PR #424, opened 2021-06-08
Add a regression test for #110

arianvp:
After the replacement node has come up, we expect the
.status.nodeStatuses to reflect the HostID of the replacement node.

Otherwise subsequent operations (like node replace) that depend on the
HostID to be correct will fail

I tried running the test suite locally using `kind`.  However the tests time out. Hoping that this PR might trigger some CI run to see the regression test fail


arianvp:
Oops. opened on the wrong repository

